PR #23, opened 2013-03-07
Remove all child nodes try 2

arv:
Compared to the previous patch this now ensures that we clear/update the internal pointers in all cases.


arv:
Let me join those two commits

sjmiles:
LbetterthistimeTM

PR #25, opened 2013-03-08
Move visual DOM manipulation into RenderShadow...

arv:
...because only the shadow renderer needs to be able to do this.

This just moves code around


sjmiles:
GTG

PR #26, opened 2013-03-08
Make offset*, client* and scroll* first flush the pending render queue

arv:


sjmiles:
Does this make changes to those values preform synchronous re-distribution, and if so, is it to avoid developer confusion over these properties having async side effects? Do we have a performance concern here?

arv:
Forget about shadow dom for a second. These properties have sync reflow because the render tree needs to be up to date to report correct layout information. OK, now you can recall shadow dom again... to report the correct layout information we must not have any pending layouts.

sjmiles:
Thank you for the info. The critical word in the request is "first".

PR #31, opened 2013-03-18
Add WrapperEventTarget and WrapperEvent

arv:
This makes EventTarget the base of Node as speced and implemented in latest Firefox.


sjmiles:
LGTM

PR #34, opened 2013-03-18
Add support for HTMLTemplateElement

arv:
For browsers that has native support for template elements we just need to
ensure that content is wrapped. For other browsers we move the template content
into a document fragment as required by the template spec.


sjmiles:
Do we need a standalone Template polyfill too? For example, I know MDV has some Template compatibility code too.

sjmiles:
LGTM

PR #36, opened 2013-03-20
Implement more NodeList methods

arv:
We do not support live NodeLists. All node lists are snap shots at the time of
the method call (or getter invokation for childNodes).


sjmiles:
LGTM

PR #37, opened 2013-03-20
Implement custom event dispatching that walks the adjusted ancestor tree

arv:
This works by listening to the underlying event using capturing. When the event
occurs we gather the adjusted ancestors and manually dispatch along that path.

This does not yet handle stopPropagation nor stopImmediatePropagation.

### DIFF
--- a/src/wrappers/Event.js
+++ b/src/wrappers/Event.js
@@ -17,8 +17,12 @@
     this.node = original;
   }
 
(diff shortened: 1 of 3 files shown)


sjmiles:
This is complicated, my review was cursory. But I believe this is the right tack, so I'm going to pull it in and we can evaluate the impact in the field.

PR #38, opened 2013-03-21
Finish event dispatching

arv:
This implementes stopPropagation, stopImmediatePropagation and fixes issues
with AT_TARGET.


arv:
@sjmiles @sorvell

src/wrappers/EventTarget.js
```diff
@@ -103,72 +105,88 @@
...
-        break;
-      var currentTarget = ancestorChain[i].ancestor;
-      var target = ancestorChain[i].target;
-      if (!deliverEvent(target, currentTarget, event, type, phase))
-        break;
+    if (!propagationStopped) {
+      propagationStopped = invoke(ancestorChain[0], event, type, phase);
+      i++;
+    }
+
```
dfreedm, line 121: Non bubbling events like `focus` need to fire in `AT_TARGET` phase on the ShadowRoot's host, which is `ancestorChain[length-1]` right?

arv, line 121: The target is at `ancestorChain[0]`.

I started implementing retargetting. I'll get to 'focus' eventually.

dfreedm, line 121: SGTM, I'll pull this now then

PR #40, opened 2013-03-21
Add a method to reset the node wrapper internal pointers.

arv:


arv:
@sorvell

sjmiles:
LGTM. Thanks for putting this in, even if it is a stopgap.

PR #44, opened 2013-03-25
Add support for document.addEventListener

arv:
This also fixes an issue where removeEventListener left a wrapped listener.


sjmiles:
LGTM

PR #45, opened 2013-03-25
Don't export to global object

arv:
Instead we export to a ShadowDOMPolyfill object.


sjmiles:
Looks good. FYI, we will probably re-export some things to 'window' at least in the interim, because we use them all over the place (mainly: wrap/unwrap). Over time, hopefully we will reduce those usages; I'm aiming to sugar the 80% cases.

PR #47, opened 2013-03-26
Remove resetNodePointers

arv:
resetNodePointers is no longer needed and in the spirit of keeping things small
it is therefore removed.


sjmiles:
Good. Sorry about the thrash!

PR #49, opened 2013-03-26
Update event.relatedTarget as needed.

arv:
This also contains fixes for the following:
1. Allow listening to event on the shadow root by addind the real event listener
   to the host.
2. If, after retargetting, we end up with target === currentTarget we need to
   change the phase during capture and ignore during bubbling.


arv:
Rebased

dfreedm:
LGTM, thanks for rebase :+1:

PR #52, opened 2013-03-27
Override builtin native constructors.

arv:
Now you can do `element instanceof HTMLElement` and it will still work.

### DIFF
--- a/src/wrappers.js
+++ b/src/wrappers.js
@@ -23,6 +23,22 @@ var ShadowDOMPolyfill = {};
     return to;
   };
 
+  function mixinStatics(to, from) {
+    Object.getOwnPropertyNames(from).forEach(function(name) {
+      switch (name) {
+        case 'arguments':
+        case 'caller':
+        case 'length':
+        case 'name':
+        case 'prototype':
+          return;
+      }
+      Object.defineProperty(to, name,
+                            Object.getOwnPropertyDescriptor(from, name));
+    });
+    return to;
+  };
+
   function getWrapperConstructor(node) {
     var nativePrototype = node.__proto__ || Object.getPrototypeOf(node);
     var wrapperConstructor = constructorTable.get(nativePrototype);
@@ -111,6 +127,7 @@ var ShadowDOMPolyfill = {};
   function register(nativeConstructor, wrapperConstructor, opt_instance) {
     var nativePrototype = nativeConstructor.prototype;
     registerInternal(nativePrototype, wrapperConstructor, opt_instance);
+    mixinStatics(wrapperConstructor, nativeConstructor);
   }
 
   function registerInternal(nativePrototype, wrapperConstructor, opt_instance) {
@@ -120,6 +137,11 @@ var ShadowDOMPolyfill = {};
     addForwardingProperties(nativePrototype, wrapperPrototype);
     if (opt_instance)
       registerInstanceProperties(wrapperPrototype, opt_instance);
+  }
+
+  function isWrapperFor(wrapperConstructor, nativeConstructor) {
+    return constructorTable.get(nativeConstructor.prototype) ===
+        wrapperConstructor;
   }
 
   /**
@@ -153,14 +175,8 @@ var ShadowDOMPolyfill = {};
     return GeneratedWrapper;
   }
 
-  function registerHTMLElement(tagName) {
-    var element = document.createElement(tagName);
-    if (element.constructor === HTMLElement ||
-        element instanceof HTMLUnknownElement) {
-      return;
-    }
-    registerObject(element);
-  }
+  var originalNode = Node;
+  var originalEvent = Event;
 
   /**
    * Wraps a node in a WrapperNode. If there already exists a wrapper for the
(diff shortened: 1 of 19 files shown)


arv:
Scott, look at test/custom-event.js. This should make you happy.

At this point it only overrides a few of the builtins. To override all we need to do more work on startup. Is this something we want? (The constructors are not configurable, otherwise we could have used getters to do this lazily.)

src/wrappers.js
```diff
@@ -111,6 +127,7 @@ var ShadowDOMPolyfill = {};
   function register(nativeConstructor, wrapperConstructor, opt_instance) {
     var nativePrototype = nativeConstructor.prototype;
     registerInternal(nativePrototype, wrapperConstructor, opt_instance);
+    mixinStatics(wrapperConstructor, nativeConstructor);
```
sjmiles, line 130: AFAIK, none of these constructors are actually callable from JS right now. Is this future-proofing?

sjmiles:
> Scott, look at test/custom-element.js. This should make you happy.

This is great. /sjmiles happy dance. 

This commit solves problems across the board.

> At this point it only overrides a few of the builtins

Hrm, not sure what you mean. Can you give me a few more words?

sjmiles:
There are pending questions, but I really wanted this code, so I merged it, which closed the request. Do we continue discussion here?

src/wrappers.js
arv, line 130: The mixin here is to get things like Node.ELEMENT_TYPE and Event.CAPTURE_PHASE set up correctly.

arv:
Right now only the following are overriden:

CharacterData
Document
DocumentFragment
Element
Event
EventTarget
HTMLContentElement
HTMLElement
HTMLShadowElement
HTMLTemplateElement
HTMLUnknownElement
MouseEvent
Node
NodeList
ShadowRoot

src/wrappers.js
arv, line 130: There is another pending issue and that is that we need to support new MouseEvent('click') etc which doesn't fit well with the custom design.

sjmiles:
Ok. Last night, for MDV support I had to add (aka hacked into patches.js) overrides for:

  HTMLTemplateElement
  Text

src/wrappers.js
sjmiles, line 130: Toolkit is using 'new CustomEvent('...')' also. Right now, I added a simple wrapper for it in patches.js.

arv, line 130: Yeah, I'm currently working on new Event etc. I'll make sure I cover
CustomEvent (you know that CustomEvent is useless, right? It is only there
for non JS bindings that cannot add new properties to objects)

On Thu, Mar 28, 2013 at 12:22 PM, Scott J. Miles
notifications@github.comwrote:

> In src/wrappers.js:
> 
> > @@ -111,6 +127,7 @@ var ShadowDOMPolyfill = {};
> >    function register(nativeConstructor, wrapperConstructor, opt_instance) {
> >      var nativePrototype = nativeConstructor.prototype;
> >      registerInternal(nativePrototype, wrapperConstructor, opt_instance);
> > -    mixinStatics(wrapperConstructor, nativeConstructor);
> 
> Toolkit is using 'new CustomEvent('...')' also. Right now, I added a
> simple wrapper for it in patches.js.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/toolkitchen/ShadowDOM/pull/52/files#r3573062
> .

## 

erik

sjmiles, line 130: > CustomEvent is useless

:(, so you'll have to straighten me out.

We use it like this:

```
doc.body.dispatchEvent(
  new CustomEvent('WebComponentsLoaded', {bubbles: true})
```

It was less code then the old way we did that. How should we do it?

arv, line 130: That is fine. But if we designed the dom for js we would not have added
CustomEvent. You could have used Event and just set detail on that.

On Thu, Mar 28, 2013 at 12:33 PM, Scott J. Miles
notifications@github.comwrote:

> In src/wrappers.js:
> 
> > @@ -111,6 +127,7 @@ var ShadowDOMPolyfill = {};
> >    function register(nativeConstructor, wrapperConstructor, opt_instance) {
> >      var nativePrototype = nativeConstructor.prototype;
> >      registerInternal(nativePrototype, wrapperConstructor, opt_instance);
> > -    mixinStatics(wrapperConstructor, nativeConstructor);
> 
>  CustomEvent is useless
> 
>  :(, so you'll have to straighten me out.
> 
> We use it like this:
> 
> `````` javascript
> doc.body.dispatchEvent(
>   new CustomEvent('WebComponentsLoaded', {bubbles: true})
> );```
> 
> It was less code then the old way we did that. How should we do it?
> 
> —
> Reply to this email directly or view it on GitHub<https://github.com/toolkitchen/ShadowDOM/pull/52/files#r3573264>
> .
> 
> 
> 
> ``````

## 

erik

sjmiles:
We will also need overrides for anything a user might want to subclass. For example, I hit an exception on

HTMLButtonElement

PR #56, opened 2013-03-28
Fix issue with Firefox old DOM bindings

arv:
The old Firefox DOM bindings (still in use for some interfaces) lazily creates
the constructor, and if we have overridden the constructor of one of the super
interfaces we get into a state where the DOM element has both wrapper and non
wrapper constructors on its prototype chain.

This also overrides a list of known HTML element constructors.


sjmiles:
Good stuff, thanks.

PR #60, opened 2013-03-29
Chai assert.instanceOf fails because it does a toString on MouseEvent.

arv:
The reason why it fails in the polyfill is that we copy the native
`MouseEvent.toString` to `WrapperMouseEvent.toString` and that function is not
generic.

https://github.com/toolkitchen/ShadowDOM/issues/59


sjmiles:
Good

PR #62, opened 2013-03-29
Some versions of Firefox do not have addEventListener etc on Node.prototype.

arv:
Use document.addEventListener instead as the base.


arv:
https://github.com/toolkitchen/ShadowDOM/issues/61

sjmiles:
Thank you

PR #72, opened 2013-04-02
add gruntfile.js for minify task, minified source, and test runner against minified source

sjmiles:
Demonstrates #65, allows testing against `grunt-contrib-uglify`


gruntfile.js
```diff
@@ -0,0 +1,76 @@
...
+    'wrappers/Element.js',
+    'wrappers/HTMLElement.js',
+    'wrappers/HTMLContentElement.js',
+    'wrappers/HTMLShadowElement.js',
+    'wrappers/HTMLTemplateElement.js',
+    'wrappers/HTMLUnknownElement.js',
+    'wrappers/generic.js',
+    'wrappers/ShadowRoot.js',
+    'ShadowRenderer.js',
+    'wrappers/Document.js',
```
arv, line 24: This is not up to date... it should have `'wrappers/Window.js'` as well

test/index.min.html
```diff
@@ -0,0 +1,61 @@
...
+})</script>
+<script src="test.js"></script>
+<script src="paralleltrees.js"></script>
+<script src="rerender.js"></script>
+<script src="reprojection.js"></script>
+<script src="custom-element.js"></script>
+<script src="HTMLShadowElement.js"></script>
+<script src="HTMLContentElement.js"></script>
+<script src="HTMLTemplateElement.js"></script>
+<script src="Document.js"></script>
```
arv: Missing `Window.js`

arv:
Is there a better way for us to not duplicate all this code?

One thing we can do is to at least move the js in the html file into a separate js file.

sjmiles:
For the other polyfills we made the <polyfill>.js file conditionally load
<polyfill>.debug.js or <polyfill>.min.js based on flags. In general that
lets us have single HTML files.

I don't know if you want to go that far.

arv:
We can have js file that has all the known files in it... like a manifest.

gruntfile.js
sjmiles, line 24: fixed

sjmiles:
Take a look at that. I factored 95% of the source into `test.main.js` which I load from `test*.html`.

test/test.main.js
```diff
@@ -0,0 +1,56 @@
...
+    assert.equal(a[i], b[i], msg);
+  }
+}
+
+mocha.setup({
+  ui: 'tdd',
+  globals: ['console', 'getInterface']
+})
+
+modules = [
```
arv: missing var?

'use strict' would have told you so.

arv:
LGTM

Please squash before merging.

sjmiles:
superseded by squashed pull #74

PR #74, opened 2013-04-03
add gruntfile for build task, build, and test runner using build

sjmiles:
factor test guts into test.main.js, referenced from *.html test runners


sjmiles:
This is the squashed version of the previous pull request #72

sjmiles:
This doesn't merge cleanly, starting over.

PR #77, opened 2013-04-04
Fix issues with window load events.

arv:
Window load events are supposed to be dispatched on the window with the target
set to document and the currentTarget set to the window.

Fixes issue #76


sorvell:
LGTM

PR #78, opened 2013-04-04
Fix issue where getShadowTree was not prefixed with scope.

arv:
This also fixes issue where inSameTree was broken as well as returning too
early in case retargetting resulted in target and relatedTarget being the same

Fixes issue #71


sorvell:
LGTM

PR #80, opened 2013-04-04
AT_TARGET should fire during the bubbling phase

arv:
Currently when the adjusted node is the same as the target node we invoke a
AT_TARGET event during the capture phase. This leads to issues where the
ordering is confusing.

See https://www.w3.org/Bugs/Public/show_bug.cgi?id=21404 for the spec bug

This fixes #68


dfreedm:
LGTM!

PR #82, opened 2013-04-05
Fix for calculating the event parent

arv:
We need to keep track of the event parent when we reditribute insertion points
since we cannot use `node.insertionParent`.

Fixes #79


dfreedm:
LGTM

PR #86, opened 2013-04-05
Implement event constructors for Safari and IE

arv:
This implements event constructors for

Event
CustomEvent
UIEvent
MouseEvent
FocusEvent

More can easily be added as needed.

Fixes #58

### DIFF
--- a/src/wrappers/EventTarget.js
+++ b/src/wrappers/EventTarget.js
@@ -330,7 +330,7 @@
     if (type instanceof OriginalEvent)
       this.impl = type;
     else
-      this.impl = new OriginalEvent(type, options);
+      this.impl = constructEvent(OriginalEvent, 'Event', type, options);
   }
   Event.prototype = {
     get target() {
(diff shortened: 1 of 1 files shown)


src/wrappers/EventTarget.js
```diff
@@ -406,6 +406,73 @@
...
+        initDict = mixin(mixin({}, superDict), initDict);
+      }
+
+      defaultInitDicts[name] = initDict;
+    };
+
+    // The order of the default event init dictionary keys is important, the
+    // arguments to initFooEvent is derived from that.
+    configureEventConstructor('Event', {bubbles: false, cancelable: false});
+    configureEventConstructor('CustomEvent', {detail: null}, 'Event');
```
dfreedm, line 459: CustomEvent works as a constructor everywhere but IE. Maybe a separate test for `new CustomEvent` would be advisable?

arv, line 459: It is true that Event and CustomEvent works in Safari. Do you think it is worth testing these individually?

dfreedm, line 459: Eh, probably not. The initialization dictionary is pretty small for both of them.

LGTM

PR #91, opened 2013-04-09
Wrap MutationObserver and MutationRecord

arv:
Fixes #54

### DIFF
--- a/shadowdom.js
+++ b/shadowdom.js
@@ -32,6 +32,7 @@
     'ShadowRenderer.js',
     'wrappers/Document.js',
     'wrappers/Window.js',
+    'wrappers/MutationObserver.js',
     'wrappers/override-constructors.js'
   ].forEach(function(src) {
     document.write('<script src="' + base + src + '"></script>');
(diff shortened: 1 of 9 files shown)


arv:
This should allow you to retire https://github.com/toolkitchen/PointerEvents/tree/shadow-dom-polyfill

dfreedm:
I'm going to test with my use in PointerEvents:

```
observe(document | shadow root,
  {attributes: true, attributeFilter: [...], childList: true, subtree: true}
)
```

arv:
That reminds me. I should probably special case `document` since we might get an unwrapped document. I can do that in this PR or in a new one. Either way is fine by me.

dfreedm:
I was just going to mention that actually :+1: 
Would it be ok to do it in this one? It's the only case I had left to wrap.

Also, maybe just one or two tests with observer watching attributes change.

arv:
There is a test watching attribute changes already. It is not as interesting because it does not need any special code.

arv:
PTAL

dfreedm:
Ah my eyes missed that test :(

LGTM!

PR #99, opened 2013-04-10
Fix issue with document.body.innerHTML

arv:
It used to end up calling `document.body.removeChild(realNode)` which does not
work because we have hijacked document.body.removeChild to only work with
wrappers.

Fixes #97


sjmiles:
LGTM

PR #102, opened 2013-04-10
Hard bind window to the window overrides

arv:
This allows us to do:

  addEventListener('click', func)

without (window)

Fixes #101

### DIFF
--- a/src/wrappers/Window.js
+++ b/src/wrappers/Window.js
@@ -10,7 +10,6 @@
   var registerWrapper = scope.registerWrapper;
   var unwrap = scope.unwrap;
   var wrap = scope.wrap;
-  var wrapEventTargetMethods = scope.wrapEventTargetMethods;
 
   var OriginalWindow = window.Window;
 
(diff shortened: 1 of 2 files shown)


sjmiles:
Is this better than just changing

```
var w = wrap(this);
```

to 

```
var w = wrap(this || window);
```

?

arv:
Less runtime checks?

What is the benefit of doing `this || window`? We should only get here if they call either `window.foo` or `foo`.

sjmiles:
Right, so the only things that can happen are 'this' is valued or 'this' is supposed to be 'window'.

(this || window) allows this to work for any future overrides instead of you having to map them explicitly.

It's also way less code, that's mainly why it popped into my head.

No big deal either way.

arv:
I'll add `this || wrap`. Better safe than sorry.

arv:
PTAL

sjmiles:
Ok, well there it's not going to do anything right? Only those exact methods ever hit that code.

I was referring to https://github.com/toolkitchen/ShadowDOM/blob/b743113c74ad70f1c986af18ef0364164d679bf2/src/wrappers.js#L264.

arv:
> I was referring to https://github.com/toolkitchen/ShadowDOM/blob/b743113c74ad70f1c986af18ef0364164d679bf2/src/wrappers.js#L264.

Changing that would be bad because then other things would fallback on `window` unexpectedly.

``` js
var ac = document.body.appendChild;
ac(node); 
```

That would call `window.appendChild`

sjmiles:
Good point.

sjmiles:
LGTM

PR #104, opened 2013-04-11
use MatchesSelector to implement simple localized querySelector

sjmiles:
Arv, PTAL, I don't know if this is the way you would like to do this.

References #70.

### DIFF
--- a/src/querySelector.js
+++ b/src/querySelector.js
@@ -1,0 +1,67 @@
+// Copyright 2013 The Toolkitchen Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+(function(scope) {
+  'use strict';
+  
+  // spec
+  var matches = Element.prototype.matches;
+  
+  // impls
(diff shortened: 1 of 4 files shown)


sjmiles:
Thanks for all the notes Arv, Please look at the latest revision when you get a chance.

src/querySelector.js
```diff
@@ -0,0 +1,82 @@
...
+  'use strict';
+
+  // imports
+   
+  var unwrap = scope.unwrap;
+
+  // find a MatchesSelector for this platform
+    
+  // spec
+  var matches = Element.prototype.matches;
```
arv, line 15: There is also:

https://github.com/toolkitchen/ShadowDOM/blob/master/src/ShadowRenderer.js#L225

We should get rid of one of them.

arv:
LGTM

dfreedm:
LGTM2

arv:
Fixed in 3373192d44e2b0da37b8d9d2dd006a631fa061ed

PR #106, opened 2013-04-12
Reset event parent when we redistribute a node.

arv:
This was causing the iloop in platform/workbench/menu-button.html


arv:
I did not manage to make a reduction of this to put in a test.

sorvell:
LGTM

PR #109, opened 2013-04-17
Add missing dependency for yuidocs into package.json

leostratus:
Submitting to master since I'm not sure what the process is to get something into stable, sounds like master is where the action is anyways.

Very minor housekeeping, this.


sjmiles:
This is the right place. Thanks!

PR #112, opened 2013-04-17
Blink now correctly exposes MutationRecord interface object.

arv:
Unfortunately the ShadowDOM polyfill did not handle instance properties
correctly.


arv:
This was breaking dev channel for me

dfreedm:
LGTM

PR #113, opened 2013-04-18
Revert "Reset event parent when we redistribute a node."

arv:
This reverts commit e701c2e92632386ccb1a4e886d2884c4e7efa0e2.

Turns out to break redistribution. Will have to do more investigation.

Fixes https://github.com/toolkitchen/toolkit/issues/102


dfreedm:
Thanks for the massive investigation :+1: 
LGTM

PR #114, opened 2013-04-19
Fix issue with event path in case of reprojection.

arv:
This fixes the issues we saw with https://github.com/toolkitchen/ShadowDOM/commit/e701c2e


arv:
@azakus 

This passes both yt-app, the workbench sample as well as all the tests. :beers:

sorvell:
LGTM.

Going to open another issue about a failing toolkit test that may be related but I think we're almost there!

PR #117, opened 2013-04-23
Rerender reprojection

arv:
When we are recursively rendering a shadow host we need to ensure it is dirty to render it again. This is because reprojection may be in effect.

Fixes #115
Closes #116


arv:
This is screwed up... let me see whatis going on

arv:
Wrong branch. Closing and creating a new PR.

PR #120, opened 2013-04-24
Fix issue with event parents

arv:
Turns out we cannot store the event parents on the insertion point, since the
insertion point might have different event parent depending on the where the
event originated from. Instead we store the event parents on the distributed
node as an array and shift off of that when we get the event path during
dispatch.

Fixes #119


arv:
I'm finally starting to understand this :smile:

sorvell:
LGTM

PR #121, opened 2013-04-25
Query selector

arv:
This is based on https://github.com/toolkitchen/ShadowDOM/tree/query-selecto

It also implements `getElementsByTagName`, `getElementsByTagNameNS` and `getElementsByClassName`.

It also installs `matches` on `Element` so we only have to do the feature detection in a single place.

### DIFF
--- a/src/wrappers/Document.js
+++ b/src/wrappers/Document.js
@@ -5,7 +5,9 @@
 (function(scope) {
   'use strict';
 
+  var GetElementsByInterface = scope.GetElementsByInterface;
   var ParentNodeInterface = scope.ParentNodeInterface;
+  var SelectorsInterface = scope.SelectorsInterface;
   var Node = scope.wrappers.Node;
   var defineWrapGetter = scope.defineWrapGetter;
   var mixin = scope.mixin;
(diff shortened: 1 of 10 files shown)


src/wrappers/Document.js
```diff
@@ -32,17 +34,62 @@
...
-        'removeChild'
-      ]);
+  forwardMethodsToWrapper([
+    window.HTMLBodyElement,
+    window.HTMLDocument || window.Document,  // Gecko adds these to HTMLDocument
+    window.HTMLHeadElement,
+  ], [
+    'adoptNode',
+    'appendChild',
+    'createDocumentFragment',
```
arv: This is not right... these should only be on document... fixing.

arv:
I'll squash these when you are happy with the changes.

sjmiles:
LGTM, thanks for tackling (and improving) this code.

arv:
I'll squash and merge

On Thu, Apr 25, 2013 at 1:49 PM, Scott J. Miles notifications@github.comwrote:

> LGTM, thanks for tackling (and improving) this code.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/toolkitchen/ShadowDOM/pull/121#issuecomment-17039442
> .

## 

erik

PR #123, opened 2013-04-26
Implement elementFromPoint for document and ShadowRoot

arv:
This uses the algorithm from the ShadowDOM spec to ensure that elementFromPoint enforces the upper boundary encapsulation.

Fixes #122


dfreedm:
LGTM!

PR #125, opened 2013-04-26
Wrap document.write

arv:
When we do a document.write we reset the lastChild_ and nextSibling_ pointers
so that next time we get lastChild or nextSibling we recompute them.

This also adds a dependency on tooltkitchen/tools to reuse the htmlSuite and
htmlTest.

### DIFF
--- a/src/wrappers/Document.js
+++ b/src/wrappers/Document.js
@@ -56,6 +56,7 @@
   ].forEach(wrapMethod);
 
   var originalAdoptNode = document.adoptNode;
+  var originalWrite = document.write;
 
   mixin(Document.prototype, {
     adoptNode: function(node) {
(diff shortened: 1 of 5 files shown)


src/wrappers/Document.js
```diff
@@ -64,6 +65,17 @@
     },
     elementFromPoint: function(x, y) {
       return elementFromPoint(this, this, x, y);
+    },
+    write: function(s) {
```
dfreedm, line 69: write is actually defined on HTMLDocument, and overrides the definition you place on Document.

arv, line 69: This area is a bit hairy. The spec only has Document (HTMLDocument is no more) so the wrappers only have Document. No matter oif the native document instance is an HTMLDocument or Document the wrapper we create is always Document.

https://github.com/arv/ShadowDOM/blob/07be1f387f9a099bd9a13a0edf193acbf0f5c522/src/wrappers/Document.js#L133

``` js
  // Both WebKit and Gecko uses HTMLDocument for document. HTML5/DOM only has
  // one Document interface and IE implements the standard correctly.
```

test/document-write.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<script src="../node_modules/chai/chai.js"></script>
+<script src="../tools/test/htmltest.js"></script>
+<script src="../shadowdom.js"></script>
+<a></a><script>
+var assert = chai.assert;
+
+var a = document.querySelector('a');
+document.write('<b></b>');
+assert.equal(a.nextSibling.localName, 'script');
```
dfreedm, line 12: This test doesn't expose the flaw.
`document.write` another `script` before `b` and then `a.nextSibling.nextSibling.localName` is `null`.

PR #128, opened 2013-04-26
Make render invalidation smarter.

arv:
We only need to invalidate if:
- This is a shadow root
- This is a shadow host
- We are changing the select property
- We are a direct child of a shadow root (this is a bit too agressive)
- We are a direct child of a shadow host


arv:
@sorvell, @sjmiles Did I miss any case?

sorvell:
LGTM, thanks!

PR #129, opened 2013-04-27
Force rendering before we dispatch any event

arv:
This is so that the event path will be correct when the event is dispatched.


sjmiles:
LGTM

PR #135, opened 2013-05-08
Wrap compareDocumentPosition

arv:
This only wraps the native implementation so it only operates on the composed DOM.

In most cases `contains` is what we want anyway.

Fixes #134


dfreedm:
LGTM

PR #142, opened 2013-05-10
MutationObserver for IE. Second try.

arv:
This time the MutationObserver polyfill is added after the shadow dom polyfill
and the mutation observer only works with wrapped nodes. It still needs to be
aware of the wrappers since observe can be called with non wrapped nodes.


dfreedm:
LGTM

PR #147, opened 2013-05-13
Add support for onfoo properties

arv:
Wrap all onfoo properties so that the function is called with a wrapped event
object.

Fixes #146


arv:
Ping

dfreedm:
LGTM, sorry for delay.

PR #151, opened 2013-05-16
Fix HTMLLegendElement for FF21

arv:
In FF21 legend uses the old XPCOM binding which has issues that changing
window.HTLMElement changes any sub class that has HTMLElement in its prototype
chain. We therefore need to ensure that we do not override HTMLElement before
overriding HTMLLegendElement.

Fixes #149


dfreedm:
LGTM

PR #160, opened 2013-06-12
Implement event.path

arv:
Fixes #153


arv:
@dglazkov This implementation diverges quite a lot from the spec. I believe the result is the same since it is using the exact same path as the event dispatching is using.

sorvell:
LGTM

PR #168, opened 2013-06-14
Adopt shadow roots

arv:
This now adopts shadow roots in the subtree as well as the older shadow roots in case there are multiple.


arv:
@azakus

dfreedm:
This DOM construction would still fail:

``` html
<div>
  <shadow-root>
    <div>
      <shadow-root></shadow-root>
    </div>
  </shadow-root>
</div>
```

The inner shadow-root still has the wrong owner document.

dfreedm:
LGTM!

PR #174, opened 2013-06-18
Fix Firefox 24 issue related to FocusEvent

arv:
FF24 does not support `document.createEvent('FocusEvent')` and to create a new
FocusEvent we need to use `new FocusEvent('x')`.

IE on the other hand does not support event constructors so we have to branch
here.

Fixes #159


dfreedm:
LGTM

PR #175, opened 2013-06-18
Disable defaultPrevented test in IE

arv:
defaultPrevented always returns false outside the event listener in IE.

https://connect.microsoft.com/IE/feedback/details/790389/event-defaultprevented-returns-false-after-preventdefault-was-called


dfreedm:
LGTM

PR #176, opened 2013-06-19
Fix issue with events and non Element/Document nodes

arv:
Fixes https://github.com/Polymer/MutationObservers/issues/6


dfreedm:
LGTM

dfreedm:
Slightly incomplete: TextNode does not have event listener wrapping.

dfreedm:
NVM, the cache was stale in the IE test VM. Really LGTM

PR #193, opened 2013-08-02
Wrap the form property

arv:
Ensure that we wrap the `HTMLFormElement` for all the `form` properties.

Fixes #189


dfreedm:
LGTM

PR #198, opened 2013-08-06
Implement wrapping of document.register

arv:
This works by creating a new object hierarchy that mimics the one passed into document.register. The returned function is a wrapper function which uses the passed in prototype as its prototype.

Fixes #190


arv:
This doesn't handle the lifecycle callbacks. I'm adding that in another PR since it is just additive anyway.

PR #201, opened 2013-08-08
Add support for deeper object structures in document.register

arv:
We now walk the prototype chain until an object that has a native wrapper is found.

Fixes #190


sorvell:
LGTM, thx!

PR #204, opened 2013-08-09
Fix issue with owner document when inserting across documents

arv:
Turns out that forcing a render does not solve the whole problem.

The main issue is that when we do an insert of a node from a different document we need to propagate the `ownerDocument` through the shadow roots.

Fixes issue #200


arv:
Ping

sorvell:
LGTM

PR #205, opened 2013-08-09
Extract template content in the wrapper

arv:
Fixes a TODO in the code. Now, there should be no way to be able to see the non extracted content.


arv:
Ping

arv:
Let's try this again.

PR #208, opened 2013-08-13
ShadowDOM workarounds

jmesserly:
two workarounds we needed for the Dart version of TodoMVC to work. Issue #200 will likely be fixed soon, and I need to follow up about the dispatchEvent issue (it may only matter for our old rendertree tests).


jmesserly:
oops. Pull request defaulted to wrong repository. Sorry!!!

PR #209, opened 2013-08-13
Make attribute dirtying smarter

arv:
We now keep track of what attributes the content select depends on and when an attribute is changed we check if the current attribute change would affect one of the attributes the renderer depends on.

Fixes #207

### DIFF
--- a/src/ShadowRenderer.js
+++ b/src/ShadowRenderer.js
@@ -164,33 +164,6 @@
     }
   }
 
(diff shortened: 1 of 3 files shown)


arv:
Right now the only reflected bindings it handles is `id` and `className`.

More will be needed but I'm not sure we should add them preemptively or not?

sorvell:
LGTM, thx for the fast service!

PR #219, opened 2013-08-21
For html, head and body ignore the internal wrappers

arv:
This is another workaround for jQuery.

The underlying problem is that wrappers are not updated as more nodes are added due to the parser.

Fixes #217


arv:
I'm not happy about this hack... I'll do some brainstorming.

arv:
I think I have a solution to this

PR #220, opened 2013-08-21
Clean up invalidation code due to add or removal of children

arv:
We only need to invalidate the renderer if we are adding or removing a direct child of a host or a shadow root.

Other cases are handled by overriding `invalidateShadowRenderer`

This cleanup was part of trying to fix #216

### DIFF
--- a/src/wrappers/Element.js
+++ b/src/wrappers/Element.js
@@ -32,13 +32,10 @@
   function invalidateRendererBasedOnAttribute(element, name) {
     // Only invalidate if parent node is a shadow host.
     var p = element.parentNode;
-    if (!p)
+    if (!p || !p.shadowRoot)
       return;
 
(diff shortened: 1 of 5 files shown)


src/wrappers/Element.js
```diff
@@ -32,13 +32,10 @@
   function invalidateRendererBasedOnAttribute(element, name) {
     // Only invalidate if parent node is a shadow host.
     var p = element.parentNode;
-    if (!p)
+    if (!p || !p.shadowRoot)
       return;
 
     var renderer = scope.getRendererForHost(p);
```
arv, line 38: This was incorrect usage of `getRendererForHost`. It should only be called for a host and it will always return a renderer.

arv:
Ping

sorvell:
LGTM

PR #221, opened 2013-08-21
Use an internal pointer for ownerDocument

arv:
This depends on #220

The earlier "solution" was to let `ownerDocument` be invalid when rendering was not valid. This worked quite for a while but eventually, due to usage of template document this became unsustainable. 

Now we use an internal pointer like we do for `parentNode` which we manually keep up to date.

Fixes #216

### DIFF
--- a/src/ShadowRenderer.js
+++ b/src/ShadowRenderer.js
@@ -249,6 +249,11 @@
     this.associateNode(host);
   }
 
+  /**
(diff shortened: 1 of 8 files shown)


arv:
Please don't merge this yet

arv:
I found out what the real problem is. A new patch is coming up.

PR #223, opened 2013-08-22
Make sure we adopt the child that we are inserting

arv:
This depends on #220

Make sure we adopt the child that we are inserting, even in the case where the real dom does not have a parent it is
important to do the adoption for the ownerDocument to stay correct.

Fixes #216


arv:
One more please

sorvell:
LGTM

PR #225, opened 2013-08-26
Faster side table

arv:
Instead of using hasOwnProperty we store an entry array and compare the key.

This makes `get` about 3x faster in V8, 2x faster in JSC, 8x faster in SpiderMonkey, 2x in Chakra.
This makes `set` about 21x faster in V8, 7x faster in JSC, 22x faster in SpiderMonkey, 3x in Chakra.
http://jsperf.com/hasown/2
http://jsperf.com/hasown/3

Note that these results are micro benchmarks so they should be taken with a grain of salt.

### DIFF
--- a/src/sidetable.js
+++ b/src/sidetable.js
@@ -13,8 +13,7 @@ if (typeof WeakMap !== 'undefined' && navigator.userAgent.indexOf('Firefox/') < 
 } else {
   (function() {
     var defineProperty = Object.defineProperty;
-    var hasOwnProperty = Object.hasOwnProperty;
-    var counter = new Date().getTime() % 1e9;
+    var counter = Date.now() % 1e9;
 
     SideTable = function() {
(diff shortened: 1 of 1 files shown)


arv:
I think it is time that we share this code between repos.

One option is to rename SideTable to WeakMap and have a WeakMap repo.

dfreedm:
Doesn't this change the semantics of Sidetable.get? With this change `weakmapInstance.get(Object.create(key))` will return the value, when before it had to be on the instance of the key.

arv:
There is a check that `key === entry[0]` which handles exactly that case.

I can add a test for this case but it is silly to put that test in Polymer/ShadowDOM. If we create a new repo I would be happy to add tests.

dfreedm:
Ah gotcha. LGTM. We should discuss moving this into it's own repo in a separate thread I think.

PR #228, opened 2013-08-27
Internal pointers

arv:
This is a big change to how the internal pointers are used and when they are updated.

Previously, we always updated the internal pointers when a dom tree was mutated. Now we only update the pointers if the node would affect a shadow renderer or if it was part of render operation.

This means that we use the underlying native DOM nodes unless the node:
1. Is a shadow host
2. Is a shadow root
3. Is an HTMLContentElement
4. Is an HTMLShadowElement
5. The parent is a shadow host
6. The parent is a shadow root

Fixes #217

It also fixes #124 in a much cleaner way. The hack could be removed.

### DIFF
--- a/src/ShadowRenderer.js
+++ b/src/ShadowRenderer.js
@@ -562,6 +562,17 @@
     return getDistributedChildNodes(this);
   };
 
+  HTMLShadowElement.prototype.invalidateShadowRenderer =
+  HTMLContentElement.prototype.invalidateShadowRenderer = function(force) {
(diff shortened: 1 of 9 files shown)


arv:
@sorvell This one is pretty big and could use some extra scrutiny.

dfreedm:
@sorvell gave a verbal LGTM

LGTM2. This also fixes `document.write` + `document.head.appendChild`, :+1:

PR #233, opened 2013-08-29
Fix issues where we did not update the renderer after DOM changes

arv:
- Changing content[select] to more restrictive needs to clean
  up non matching nodes (we clear the child nodes now)
- Changes to fallback content needs to invalidate
- Adding a content or shadow needs to invalidate

Fixes #226, #235


sjmiles:
LGTM

PR #236, opened 2013-09-04
Use a diff to do minimal DOM changes during rendering

arv:
We now render the composed tree to a RenderNode tree. After that we diff that with the real DOM tree and
based on the diff we make minimal changes.

Fixes #227

### DIFF
--- a/build.json
+++ b/build.json
@@ -1,4 +1,5 @@
 [
+    "../observe-js/src/observe.js",
     "src/sidetable.js",
     "src/wrappers.js",
     "src/wrappers/events.js",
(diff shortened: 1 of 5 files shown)


build.json
```diff
@@ -1,4 +1,5 @@
 [
+    "../observe-js/src/observe.js",
```
sjmiles, line 2: Nothing wrong here, but it's worth noting the new dependency on `observe-js`. Remember that the individual polyfills can be used in isolation. IOW, theoretically, `observe-js` may need to be cloned manually.

sjmiles:
LGTM

PR #239, opened 2013-09-05
Use an expando property for the wrapper.

arv:
I did some simple benchmarks by traversing the HTML spec and a few times and this reduced the self time from 450ms to 25ms (30% to 3%). These numbers are not very exact but it is clear that this is a huge win.


arv:
I'm TBR'ing this.

PR #240, opened 2013-09-05
Use eval for default generated getters, setters and methods.

arv:
This allows engines to inline the functions which in my naive html spec test removes about 30% of the execution time.

### DIFF
--- a/src/wrappers.js
+++ b/src/wrappers.js
@@ -11,6 +11,19 @@ var ShadowDOMPolyfill = {};
   var nativePrototypeTable = new SideTable();
   var wrappers = Object.create(null);
 
+  // Don't test for eval if document has CSP securityPolicy object and we can
+  // see that eval is not supported. This avoids an error message in console
+  // even when the exception is caught
+  var hasEval = !('securityPolicy' in document) ||
+      document.securityPolicy.allowsEval;
+  if (hasEval) {
+    try {
+      var f = new Function('', 'return true;');
+      hasEval = f();
+    } catch (ex) {
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


arv:
TBR

goto100:
Can somebody explain this why use `new Function` makes better performance?

sapjax:
In my test (chrome  34.0.1847.131) , the result is contrary.   don't use `new Function` can removes about 50% of the execution time. 

these are my test files, you can look the console info:
http://sandbox.runjs.cn/show/wdayhxl8     (use eval)
http://sandbox.runjs.cn/show/bytu8neq     (not use eval)

the only difference is  code  `hasEval = false;` on line 1883  http://sandbox.runjs.cn/uploads/rs/353/v2pk6wp5/platform.concat.js

arv:
@goto100 It is faster because there are less work to do on each getter/setter.

@sapjax What are you measuring? The important thing to remember is that execution time is more important than startup time. Still, latest Chrome 36 no longer uses this code since it has native Shadow DOM and in Firefox the startup time difference is not as big (437ms vs 382ms).

sapjax:
@arv When the platform.js 'startup', the `new Function` run 2040 times (maybe something else is wrong). 
In mobile android browser, the startup will costs nearly 1300ms, and users can't bear it.
I agree "execution time is more important than startup time", but the startup time of ShadowDOM polyfill itself is also execution time.

arv:
`new Function` is run once per getter/setter. Maybe we could refactor this so that `new Function` is  only run once per interface.

PR #241, opened 2013-09-05
remove extra newline from enclosedBy in events.js

jmesserly:
Fixes a small style nit caught by @efortuna here:
https://codereview.chromium.org/23548014/diff/4001/pkg/shadow_dom/lib/shadow_dom.debug.js#newcode581


arv:
Nice

PR #242, opened 2013-09-06
Optimize rendering

arv:
Since we no longer mutate the tree as we render (only after rendering) we do not need to use snapshots of the children.

Stop using forEach since it is slow compared to normal for loops.

Use instance of instead of comparing localName.


arv:
This depends on #236

sjmiles:
LGTM

PR #247, opened 2013-09-09
Make sure we call nodeWasAdded_ on all added nodes

arv:
On the case of a document fragment being added we need to call nodeWasAdded_ on all the nodes that were added.

Fixes #246


frankiefu:
LGTM

PR #248, opened 2013-09-09
Fix grunt tests

dfreedm:


dfreedm:
#242 introduced a `observe-js` dependence, which was not reflected in the karma test server path.

PR #251, opened 2013-09-11
Make sure we do not remove the shadow dom of nested hosts.

arv:
When we rendered the outer tree and hit the inner host, its renderer was non dirty so the render tree was not created for that. When we later synced the trees it looked like the inner host should have no children.

The solution is to mark the render node as `skip` so that we do not recurse into it when we sync the trees.

Fixes #245


sorvell:
LGTM

PR #252, opened 2013-09-11
No need to associate node with renderer just to remove stale child nodes

arv:
This one is no longer needed now that we sync the dom trees


sorvell:
LGTM

PR #253, opened 2013-09-11
Change from using a WeakMap to use expando

arv:
Both the `shadowRoot` and renderer were backed by weak maps before and these get queried all the time. Changing them to expando for performance.


dfreedm:
LGTM.

PR #256, opened 2013-09-13
Call nodeWasAdded on all the nodes in the added tree.

arv:
Also, rerender all dirty shadow renderers.

Fixes #255


frankiefu:
This doesn't fix the more complex example in here: https://github.com/Polymer/ShadowDOM/issues/255#issuecomment-24428312

arv:
PTAL. I went back to the brute force for now. I have some ideas how to make this better in the future

frankiefu:
LGTM

PR #261, opened 2013-09-24
log the exception stack if available

morethanreal:


src/wrappers/events.js
```diff
@@ -336,7 +336,7 @@
         if (window.onerror)
           window.onerror(ex.message);
         else
-          console.error(ex);
+          console.error(ex.stack || ex);
```
arv: This prints less information than before. How about:

console.error(ex, ex.stack);

morethanreal: done

arv:
LGTM

PR #263, opened 2013-09-27
Fix issue with syncing render nodes

arv:
Fixes #259


arv:
This is why I should stick to standard for loops and not try to be fancy.

sorvell:
LGTM

PR #264, opened 2013-09-27
Mute mutation events during rendering and template content extraction

arv:
The extraction of template content was causing failures in browsers without native template element because it hit an iloop due to mutation events trying to create the wrapper.


sorvell:
LGTM

PR #267, opened 2013-10-10
Allow dispatching the same event multiple times

arv:
Fixes #266


arv:
@azakus

dfreedm:
LGTM, fixing merge with #269

PR #269, opened 2013-10-10
Add work around for X-Tag issue related to relatedTarget

arv:
X-Tag sets relatedTarget on a CustomEvent. If they do that there is no
way to have relatedTarget return the adjusted target but worse is that
the originalEvent might not have a relatedTarget so we hit an assert
when we try to wrap it.

Fixes #268


dfreedm:
LGTM, thx!

PR #270, opened 2013-10-10
Fix callback names for Custom Elements

dfreedm:
Spec change: {entered,left}DocumentCallback -> {entered,left}ViewCallback


arv:
LGTM

PR #274, opened 2013-10-15
Add wrapper for canvas and its contexts

arv:
This wraps `HTMLCanvasElement` so that we can return wrapped `CanvasRenderingContext2D` and `WebGLRenderingContext`. The context methods that take an `HTMLElement` are then wrapped to unwrap that argument.

Fixes #178


dfreedm:
Nit: Should there be some tests with "experimental-webgl" getContext?
Safari 6 supports that through a flag.

LGTM

arv:
Not sure if that is worth testing. It will work if no matter what since experimental-webgl returns the same interface.

PR #275, opened 2013-10-21
Use tools repo for mocha and chai

dfreedm:
Zero install browser test runner


arv:
This seems fine but what is the motivation for this? I don't think people expect to have to run tools npm install. Can we have pull-all.sh do this?

dfreedm:
Tools has a pre-checked out version of mocha and chai. That way, no one has to do `npm install` or `bower install` unless they are using grunt.

arv:
LGTM

PR #279, opened 2013-10-23
Fix issue with non identifier names

arv:
Latest nightly Firefox has `"@@iterator"` properties on some DOM elements which leads us to generate invalid code for eval.


dfreedm:
LGTM

PR #280, opened 2013-10-23
Add support for new Image

arv:
Doing `new Image` creates an instance of HTMLImageElement

We need to use `rewrap` because the img created with document.createElement is already associated with a wrapper.


dfreedm:
LGTM

PR #284, opened 2013-10-28
Fix issue with canvas parameters

arv:
texImage2D, texSubImage2D and drawImage are overloaded to accept a lot of different combinations. Update the wrappers to correctly handle all the current overloads.

Fixes #283


dfreedm:
LGTM

PR #287, opened 2013-10-30
wrappers.js: must set `hasEval` to false if Function constructor threw an exception

sjmiles:
If the `securityPolicy` test fails but `new Function()` throws an exception, we must conclude `hasEval` should be false.


sjmiles:
@arv 
@azakus

sjmiles:
I messed up the edit, second try incoming.

arv:
LGTM. Feel free to push your new commit without a PR

sjmiles:
Ok, thx Arv (proper fix in #288).

PR #289, opened 2013-10-30
Add the host property to the ShadowRoot interface

arv:
https://dvcs.w3.org/hg/webcomponents/raw-file/tip/spec/shadow/index.html#api-shadow-root-host

Fixes #286


arv:
Ping

sorvell:
LGTM

PR #291, opened 2013-10-31
IE document fragment work arounds

arv:
DocumentFragments are broken in Internet Explorer. There are cases where they are not valid trees.

These work-arounds try to minimize the damage done in these cases.


arv:
Ping

sorvell:
sad but LGTM

PR #292, opened 2013-11-01
Preserve 'extends' when the ShadowDOM polyfill patches document.register

sigmundch:
I noticed this bug only when using the shadow-dom polyfill in combination with the native document.register. Here is a sample repro example:

``` html
<!DOCTYPE html>
<html><body>
  <script src="ShadowDOM/shadowdom.js"></script>

  <x-b></x-b> <span is="x-c"></span>

  <script>
    var bp = Object.create(HTMLElement.prototype);
    bp.createdCallback = function() { console.log("b created"); };
    var B = document.register('x-b', { prototype: bp });

    var cp = Object.create(HTMLSpanElement.prototype);
    cp.createdCallback = function() { console.log("c created"); };
    var C = document.register('x-c', { prototype: cp, extends: 'span' });
  </script>
```

You'd expect to see: `b created` and `c created`, but only `b created` is printed.

### DIFF
--- a/src/wrappers/Document.js
+++ b/src/wrappers/Document.js
@@ -150,12 +150,19 @@
         };
       });
 
-      var nativeConstructor = originalRegister.call(unwrap(this), tagName,
-          {prototype: newPrototype});
+      var p = {prototype: newPrototype};
+      if (object.extends)
+        p.extends = object.extends;
+      var nativeConstructor = originalRegister.call(unwrap(this), tagName, p);
 
       function GeneratedWrapper(node) {
-        if (!node)
-          return document.createElement(tagName);
+        if (!node) {
+          if (object.extends) {
+            return document.createElement(object.extends, tagName);
+          } else {
+            return document.createElement(tagName);
+          }
+        }
         this.impl = node;
       }
       GeneratedWrapper.prototype = prototype;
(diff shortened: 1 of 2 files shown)


src/wrappers/Document.js
```diff
@@ -151,6 +151,7 @@
       });
 
       var nativeConstructor = originalRegister.call(unwrap(this), tagName,
+          object.extends ? {prototype: newPrototype, extends: object.extends} :
```
arv: How about:

``` js
var p = {prototype: newPrototype};
if (object.extends)
  p.extend  = object.extends;
```

arv:
Can you add a test too?

sigmundch:
good call - I had not realized that the constructor also had to be fixed (we don't hit that code path in our code). I added tests for both, passes on FF 25, Chrome 30, and Chrome 31 (where we used the polyfill + native document.register)

PTAL

arv:
LGTM

martyglaubitz:
Huh ? Why am i mentioned here ? Never contributed to this repo in anyway...

sigmundch:
@martyglaubitz I think when you did a merge in your clone of here https://github.com/martyglaubitz/bleeding_edge/commit/13b7346dc9cef05c7c2cc81ad7b326eccbeb0198, you included a commit that I made which mentions this pull request.

PR #294, opened 2013-11-01
Implement MutationObserver

arv:
This is an implementation of the MutationObserver without relying on native MutationEvents nor native MutationObserver.

Part of the implementation and almost all of the tests come from https://github.com/Polymer/MutationObservers


arv:
@ajklein @rafaelw You are the experts. Please review.

test/js/MutationObserver/childList.js
```diff
@@ -0,0 +1,479 @@
...
+      var observerB = new MutationObserver(function() {});
+      observerB.observe(b, {
+        childList: true
+      });
+
+      b.appendChild(df);
+
+      var recordsDf = observerDf.takeRecords();
+
+      assert.equal(recordsDf.length, 0);
```
arv: This looks wrong to me. Let me look at it next week.

arv:
Looking more into this I think there is a bug in the spec? I cannot find where it says to enqueue the removedNodes when a node is inserted into a new parent.

I've updated the implementation to match Blink, Gecko and Trident.

PTAL

arv:
Ping

arv:
Fixed in 56e3d82e63b498c5469cbcb2b0b782813cbf98d0

PR #298, opened 2013-11-05
Fix issue with missing instance properties on canvas context classes

arv:
Since Blink/WebKit uses magic instance properties instead of accessors we need to pass in an object listing
the instance properties.


dfreedm:
LGTM

PR #300, opened 2013-11-05
Don't use native cloneNode

arv:
The only way we could use the native cloneNode would be if we knew if there wasn't a shadow root in the current tree. We currently do not keep track of that state so always walk the tree when cloning.

Fixes #296


dfreedm:
LGTM

PR #306, opened 2013-11-13
Remove support for mutation events

arv:
In 40de72f5f6e2efa3192dc3bd26df37457eb7e931 we started muting mutation events during template extraction and rendering. Now that we have mutation observers working this manual muting is no longer needed.

This also removes support for MutationEvent.


arv:
Ping

sorvell:
LGTM

PR #307, opened 2013-11-13
Add way to blacklist certain instance properties

arv:
When we create the wrappers we currently iterate over all instance properties due to bugs in Blink/WebKit. If one of these instance properties are deprecated the Dev Tools logs a warning. To work around this we can now add a black list of property names that should not be transferred to the wrapper prototype.

Fixes #303


arv:
Ping

sorvell:
LGTM

PR #308, opened 2013-11-13
Fix issue with WebIDL attributes on Range

arv:
registerWrapper needs an instance object to be able to pick up instance properties.

Fixes #301


dfreedm:
LGTM

PR #311, opened 2013-11-14
Move Range.js in build.json to match what we do in shadowdom.js

sigmundch:
This is basically to match the change here https://github.com/Polymer/ShadowDOM/commit/57cb3435c64b560f9db8af3396eaa929d55285b9#diff-7350cd3f3c3f571d71cc26670bcb779dR42 for the uglified code.


dfreedm:
I _just_ ran into this, thanks!

LGTM

PR #318, opened 2013-11-21
Corrects document.importNode, fixes #317

sorvell:


arv:
LGTM

arv:
I just realized that this is not enough. We need to manually do what we do in cloneNode to ensure that we do not "leak" the shadow tree. I'll work on a patch.

PR #321, opened 2013-11-22
importNode needs to traverse the tree

arv:
Fixes #317


sorvell:
yeah, makes sense. LGTM.

PR #326, opened 2013-11-25
Add work around for Firefox bug related to window methods

arv:
https://bugzilla.mozilla.org/show_bug.cgi?id=943065

This is a regression in the nightly. Once fixed in Firefox the work around should be removed


sorvell:
LGTM

PR #329, opened 2013-12-02
Use emulated dispatchEvent for EventTarget.prototype.dispatchEvent

arv:
If the code calls dispatchEvent then we go straight to the emulated path. This is important because the node might not be connected to the node where we have listeners on in the native tree.

Fixes #327


sorvell:
LGTM

PR #330, opened 2013-12-04
new FocusEvent fails in Safari

arv:
Older versions of Safari does not support new FocusEvent but Firefox does not support createEvent('FocusEvent'). What to do? I know, lets use try/catch!


sjmiles:
LGTM

PR #331, opened 2013-12-10
Enhance compatibility with native custom elements

sorvell:
This fixes issue #190.

Elements that are wrapped prior to custom element registration have stale wrappers. This is not addressed generically, but we do address it here in the custom element callbacks such that the element has the expected wrapper in these callbacks. If an element does not have one of these callbacks and is wrapped prior to registration, the user must 'rewrap' the element via:

```
ShadowDOMPolyfill.rewrap(ShadowDOMPolyfill.unwrap(element));
```

It's not possible to fix this generically since the invalid wrapper reference is held by the user.

### DIFF
--- a/src/wrappers/Document.js
+++ b/src/wrappers/Document.js
@@ -162,7 +162,7 @@
         newPrototype[name] = function() {
           // if this element has been wrapped prior to registration,
           // the wrapper is stale; in this case rewrap
-          if (!(wrap(this) instanceof GeneratedWrapper)) {
+          if (!(wrap(this) instanceof CustomElementConstructor)) {
             rewrap(this);
           }
           f.apply(wrap(this), arguments);
@@ -173,7 +173,7 @@
       if (object.extends)
         p.extends = object.extends;
 
-      function GeneratedWrapper(node) {
+      function CustomElementConstructor(node) {
         if (!node) {
           if (object.extends) {
             return document.createElement(object.extends, tagName);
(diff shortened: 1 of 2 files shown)


src/wrappers/Document.js
```diff
@@ -159,6 +160,11 @@
         if (!f)
           return;
         newPrototype[name] = function() {
+          // if this element has been wrapped prior to registration,
+          // the wrapper is stale; in this case rewrap
+          if (!(wrap(this) instanceof GeneratedWrapper)) {
```
arv: I found this confusing at first because I though that GeneratedWrapper was the GeneratedWrapper from wrappers.js. Maybe we could rename it to CustomElementConstructor?

test/js/Document.js
```diff
@@ -439,6 +439,25 @@ htmlSuite('Document', function() {
...
+      return;
+
+    div = document.body.appendChild(document.createElement('div'));
+    div.innerHTML = '<x-a2-1></x-a2-1>';
+
+    function A() {}
+    A.prototype = {
+      __proto__: HTMLElement.prototype,
+      createdCallback: function() {
+        assert.isTrue(this.isCustom);
```
arv, line 453: Can you also add?

``` js
assert.instanceOf(this, A);
```

test/js/Document.js
```diff
@@ -439,6 +439,25 @@ htmlSuite('Document', function() {
...
+      },
+      isCustom: true
+    }
```
arv: Nit, semicolon

arv:
LGTM

Would you mind expanding the PR title and PR description a bit?

sorvell:
Merging via previous review, thanks.

PR #332, opened 2013-12-10
Avoid stack overflow when Element.matches is supported by the browser

foolip:
In http://crbug.com/326652 support for Element.matches(selectors) was
added to Blink, but it was reverted because it caused tests using
Polymer to start failing due to "Maximum call stack size
exceeded".

The problem is that the Element wrapper unconditionally implements
Element.prototype[matchesName] using this.matches(), which is the same
function if matchesName == "matches". Make the forwarding function
conditional to solve the problem.

### DIFF
--- a/src/wrappers/Element.js
+++ b/src/wrappers/Element.js
@@ -88,9 +88,11 @@
     }
   });
 
-  Element.prototype[matchesName] = function(selector) {
-    return this.matches(selector);
-  };
+  if (matchesName != "matches") {
+    Element.prototype[matchesName] = function(selector) {
+      return this.matches(selector);
+    };
+  }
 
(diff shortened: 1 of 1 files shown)


foolip:
Why does Element.prototype[matchesName] need to be exposed at all? If the code could just be removed altogether, that would be an even nicer fix.

Also, what will be the process for getting this into Chromium so that I can land Element.matches(selectors)?

arv:
LGTM

Element.prototype[matchesName] is needed in the case where people use mozMatchesSelector.

I'll find the people maintaining the copy of this in chromium.

foolip:
I probably don't understand how Polymer works, but why aren't all the other prefixed attributes, like webkitRequestFullScreen, similarly exposed?

arv:
Because they are not used at this point.

PR #337, opened 2013-12-17
Dispatch native events if possible

arv:
We now check whether we can dispatch a native event or if we need to
fallback on fully emulated dispatch.

Fixes #335

### DIFF
--- a/src/wrappers/events.js
+++ b/src/wrappers/events.js
@@ -656,9 +656,61 @@
       }
     },
     dispatchEvent: function(event) {
-      dispatchEvent(event, this);
+      // We want to use the native dispatchEvent because it triggers the default
+      // actions (like checking a checkbox). However, if there are no listeners
+      // in the composed tree then there are no events that will trigger and
+      // listeners in the non composed tree that are part of the event path are
+      // not notified.
+      //
+      // If we find out that there are no listeners in the composed tree we add
+      // a temporary listener to the target which makes us get called back even
+      // in that case.
+
+      var nativeEvent = unwrap(event);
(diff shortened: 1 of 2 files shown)


arv:
@jmesserly

jmesserly:
To check my understanding: if a listener is registered (hasListenerInAncestors), it will dispatch natively triggering default action (fixing #335), then the polyfill gets the event and gets a chance to simulate correct child->host dispatch (meaning #327 is still fixed too). If we don't have any listeners (neither hasListenerInAncestors nor hasListenerInEventPath), it will dispatch natively (fixing #335, and #327 doesn't really apply).

What happens in the (!hasListenerInAncestors && hasListenerInEventPath)? In that case, will the default action still be triggered?

arv:
> What happens in the (!hasListenerInAncestors && hasListenerInEventPath)? In that case, will the default action still be triggered?

In that case the default action will not be triggered.

jmesserly:
Makes sense! Well, LGTM in that it fixes many cases of #335 while not regressing #327 :)

For the remaining case: could we register a listener on the node we're dispatching on, then remove it immediately as soon as the event dispatches? It seems like that would covert the hasListenerInEventPath into the hasListenerInAncestors case (and also makes dispatchEvent more similar to a user action triggering the event).

arv:
That seems like a good idea. Let me try it.

arv:
Thanks @jmesserly. That turned out to work perfectly.

PTAL

jmesserly:
Lovely! LGTM

PR #339, opened 2014-01-06
Update support for CustomElements to use document.registerElement

sorvell:
change document.register to document.registerElement
change entered/leftViewCallbacks to attached/detachedCallbacks


arv:
LGTM

PR #341, opened 2014-01-07
Fix issue with getElementById

arv:
`querySelector('#id')` only allows valid CSS IDs but `getElementById` should allow any character but "HTML space" characters.

### DIFF
--- a/src/wrappers/ShadowRoot.js
+++ b/src/wrappers/ShadowRoot.js
@@ -16,6 +16,8 @@
   var shadowHostTable = new WeakMap();
   var nextOlderShadowTreeTable = new WeakMap();
 
+  var spaceCharRe = /[ \t\n\r\f]/;
+
   function ShadowRoot(hostWrapper) {
     var node = unwrap(hostWrapper.impl.ownerDocument.createDocumentFragment());
     DocumentFragment.call(this, node);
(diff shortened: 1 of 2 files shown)


arv:
@jmesserly

src/wrappers/ShadowRoot.js
```diff
@@ -56,7 +58,9 @@
     },
 
     getElementById: function(id) {
-      return this.querySelector('#' + id);
+      if (spaceCharRe.test(id))
+        return null;
+      return this.querySelector('[id="' + id + '"]');
```
jmesserly, line 63: I think there's a corner case here with double quotes in the ID. Probably no one will ever hit it though since quotes in IDs is crazy :)
Whether it's worth the cost of escaping to fix, I'm not sure. Probably not.

``` js
# at a JS console
var d = document.createElement('div')
d.innerHTML = '<div id=\'4"2\'>'
document.body.appendChild(d)
document.querySelector('[id="4"2"]')
# SyntaxError: Failed to execute query: '[id="4"2"]' is not a valid selector.
document.querySelector('[id="4\\"2"]')
# returns: <div id=​"4"2">​</div>​
```

jmesserly:
LGTM!

src/wrappers/ShadowRoot.js
arv, line 63: I thought of this too but decide not to support it. We can always add a `replace(/"/g, '\\\\"')` later.

PR #343, opened 2014-01-10
Fix issue with SVGElementInstance

### LINKED ISSUE #388: Assertion Error thrown
I have the following component, which creates an SVG three line icon used for menus in modern UIs:

```
<link rel="import" href="../polymer/polymer.html">

<polymer-element name="nav-button">
  <template>
    <style>
      :host {
        display: block;
        background-color: red;
      }
      :host #icon line {
        stroke: rgb(255, 255, 255);
        stroke-linecap: round;
        stroke-width: 4;
      }
      :host #icon defs line {
        stroke: rgb(255, 255, 255);
        stroke-linecap: round;
        stroke-width: 4;
      }
    </style>
    <svg id="icon" version="1.1" width="32px" height="32px" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32">
      <defs>
        <line id="line" x1="0" y1="0" x2="18" y2="0" />
      </defs>
      <use xlink:href="#line" x="6" y="8" />
      <line id="middle" x1="6" y1="16" x2="24" y2="16" />
      <line id="bottom" x1="6" y1="24" x2="24" y2="24" />
    </svg>
  </template>
  <script>
    (function(undefined) {
      'use strict';
      Polymer('nav-button', {
        ready: function() {
        }
      });
    }())
  </script>
</polymer-element>
```

When hovering over the first line (the one that is instantiated with the `<use>` SVG element, Errors are repeatedly thrown from `wrappers.js:30`, with the following callstack:

`wrapper.js:30`:

```
function assert(b) {
  if (!b)
    throw new Error('Assertion failed');
};
```

`wrapper.js:290`:

```
function wrap(impl) {
  if (impl === null)
    return null;

  assert(isNative(impl));
  return impl.polymerWrapper_ ||
      (impl.polymerWrapper_ = new (getWrapperConstructor(impl))(impl));
}
```

`event.js:190`:

```
function dispatchOriginalEvent(originalEvent) {
  // Make sure this event is only dispatched once.
  if (handledEventsTable.get(originalEvent))
    return;
  handledEventsTable.set(originalEvent, true);

  return dispatchEvent(wrap(originalEvent), wrap(originalEvent.target));
}
```

This is solved by adding the `noscript` tag to the Polymer element or replacing the lines with direct instantiation (`<line>`).

Is Polymer scanning the `<use>` and digging out the `href` for something?

arv:
Implements the SVGElementInstance interface and uses fixes issues trying to wrap native SVGElementInstance objects.

Fixes https://github.com/Polymer/polymer/issues/388


dfreedm:
Looks like a few tests are failing with this PR on Safari 7.
Error seems to stem from SVGUseElement having a readonly instanceRoot property?

arv:
I guess I have to unpack my Mac after the move in December after all :'(

arv:
PTAL

I fixed the Safari issue by following the common pattern of creating the constructor/prototype first and filling it in with data from the native instance.

dfreedm:
LGTM!

PR #344, opened 2014-01-10
Blacklist keyLocation from all Events

arv:
Chrome warns about keyLocation.

KeyboardEvent is a bit buggy in Firefox so I decided to just put add the name to the Event blacklist.


dfreedm:
LGTM

PR #348, opened 2014-01-14
Fix issues with innerHTML in plaintext like elements

arv:
This also cleans up innerHTML escaping for attributes, text nodes
and comments.

Fixes #347


arv:
@blois

arv:
@azakus Do you mind taking a look?

dfreedm:
LGTM

PR #350, opened 2014-01-15
Fix issue with Element not having all the required properties

arv:
Fixes #349


arv:
@rafaelw The IE issue should not affect your svg:template patch since the namespaceURI stays correct even though the interface is not correct.

arv:
Ping

dfreedm:
LGTM!

dfreedm:
Ah nuts, getting some failed tests on Safari with this patch.

PR #351, opened 2014-01-15
Fix element issue again

arv:
The last one had a typo in it. It should have been using createElementNS in the first place.

Also, disable the `instanceof SVGElement` for svg/template since IE11 and Safari7 both break on it.


arv:
@azakus TBR

PR #352, opened 2014-01-15
Fix issue related to matches and jQuery

arv:
Chrome Canary now ships with `Element.prototype.matches` so we ended up only forwarding `document.documentElement.matches` to the wrapper. With this patch we forward all of them (matches and webkitMatchesSelector/mozMatchesSelector/msMatchesSelector).


dfreedm:
LGTM!

PR #356, opened 2014-01-16
Fix issue where the DOM tree could get out of sync.

arv:
The problem occurs when a node that requires invalidation is moved into
anohter node that does not require invalidation. The internal pointers
are no updated as expected.

Fixes https://github.com/Polymer/platform/issues/48

### DIFF
--- a/src/wrappers/Node.js
+++ b/src/wrappers/Node.js
@@ -184,6 +184,18 @@
     return df;
   }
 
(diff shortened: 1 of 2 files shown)


arv:
I'm hoping I can find a cleaner solution but this fixes the issue for now.

dfreedm:
LGTM

PR #360, opened 2014-01-22
Add Selection interface

arv:
Fixes #359


dfreedm:
Range toString fails on Safari 7

arv:
Oh, Safari... looking

arv:
Fixed by making sure the div is in the document when we select it using range selectNode

dfreedm:
LGTM!

PR #362, opened 2014-01-23
Fix IE9 issue with shadow tree composition

arv:
msMatchesSelector does not work on elements that are not in the document


dfreedm:
LGTM

PR #363, opened 2014-01-23
Fix IE9 issue with innerHTML on plaintext elements

arv:
Setting innerHTML on plaintext elements creates non Text children.
Check if the element is a plaintext elemend and if so use textContent
instead.


dfreedm:
LGTM

PR #368, opened 2014-01-29
Text normalize

dfreedm:
Node.normalize concatenates Text nodes, only splitting them when other structural nodes are between.

Without this patch, a node with a ShadowRoot will accidentally process the ShadowRoot's text nodes.

### DIFF
--- a/src/wrappers/Node.js
+++ b/src/wrappers/Node.js
@@ -228,6 +228,13 @@
     return p && p.invalidateShadowRenderer();
   }
 
+  function cleanupNodes(nodes) {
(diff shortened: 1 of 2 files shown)


dfreedm:
@arv, will squash before merging.

dfreedm:
The spec also mentions live updating any Range that includes the affected Text Nodes. Seems like it might be a significant performance penalty to track all Ranges and update them. Thoughts?

dfreedm:
@arv PTAL

arv:
LGTM

Updating the Ranges is something I've skipped before.

PR #370, opened 2014-02-03
innerHTML was not working correctly for HTMLTemplateElement

arv:
The logic for handling innerHTML (and outerHTML) for template elements needs to be handled in the generic innerHTML function due to the recursive call.


dfreedm:
LGTM

PR #371, opened 2014-02-03
Template element clone node

arv:
Fix issues with cloneNode and importNode for HTMLTemplateElement

Refactor the code to share the implementation.
#367

### DIFF
--- a/src/wrappers/Node.js
+++ b/src/wrappers/Node.js
@@ -17,6 +17,7 @@
   var unwrap = scope.unwrap;
   var wrap = scope.wrap;
   var wrapIfNeeded = scope.wrapIfNeeded;
+  var wrappers = scope.wrappers;
 
   function assertIsNodeWrapper(node) {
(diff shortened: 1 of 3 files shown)


src/wrappers/Node.js
```diff
@@ -235,6 +236,34 @@
     }
   }
 
+  var originalImportNode = document.importNode;
+  var originalCloneNode = Node.prototype.cloneNode;
+
+  function cloneNode(node, deep, opt_doc) {
+    var clone;
+    if (opt_doc)
```
dfreedm, line 244: It looks like `opt_doc` is always defined, since both `cloneNode` and `importNode` pass one.

arv, line 244: My intent was to not pass a doc for cloneNode. Let me fix that.

arv:
Done. PTAL

dfreedm:
LGTM

PR #373, opened 2014-02-04
Workaround to don't crash when loading in Opera 12

gartz:
This lite fix make it load without errors in the Opera 12


gartz:
Added suport for Opera native matches

arv:
At this point we do not see a need to support Opera 12. Our goal is to get rid of the polyfills eventually and Opera 12 is a dead end.

My recommendation would be to maintain a fork for now.

PR #376, opened 2014-02-12
Always return an array for getDisributedNodes

arv:
Partial work around for https://github.com/Polymer/platform/issues/52


dfreedm:
LGTM

PR #377, opened 2014-02-20
Add support for HTMLSelectElement add and remove

arv:
Since HTMLSelectElement add takes a Node we need to manually unwrap it as needed.


dfreedm:
LGTM

PR #378, opened 2014-02-20
Change override constucturs to filter out wrappers

arv:
This is mostly cleanup but it does remove some unnecessary createElement at startup and it does provide better test coverage.


dfreedm:
LGTM

PR #380, opened 2014-02-21
Implement the HTML Table DOM APIs

arv:
This implements the following interfaces
- HTMLTableElement
- HTMLTableSectionElement
- HTMLTableRowElement

I punted on HTMLCollection it is right now just an alias to NodeList.
HTMLTableDataCellElement, HTMLTableCellElement etc have no methods that return nodes so there is no need to manually implement these wrappers.

Fixes #366


arv:
Ping

dfreedm:
LGTM!

PR #384, opened 2014-02-26
Rendering needs to invalidate nested shadow trees

arv:
If we don't do this the redistribution breaks.

Fixes #383


arv:
I'm looking into ordering the rendering from top to bottom now so we do not have to render things more than once.

sorvell:
LGTM

PR #386, opened 2014-02-27
Set the prototype after all the mutations to allow optimizations

arv:
Some VMs optimize objects when they are set as a prototype.


dfreedm:
LGTM!

PR #388, opened 2014-02-27
Introduce a TreeScope so we can get to the root node in O(1)

arv:


arv:
This is the better solution for #383. We now ensure that we do not render descendants before ancestors.

sorvell:
LGTM

PR #391, opened 2014-03-03
Rename the .impl property and ensure all files use unwrap as needed

arv:
This is in preparation of allowing different strategies for wrapping

### DIFF
--- a/src/wrappers.js
+++ b/src/wrappers.js
@@ -136,23 +136,31 @@ window.ShadowDOMPolyfill = {};
     return /^\w[a-zA-Z_0-9]*$/.test(name);
   }
 
+  // The name of the implementation property is intentionally hard to
+  // remember. Unfortunately, browsers are slower doing obj[expr] than
+  // obj.foo so we resort to repeat this ugly name. This ugly name is never
+  // used outside of this file though.
+
   function getGetter(name) {
     return hasEval && isIdentifierName(name) ?
-        new Function('return this.impl.' + name) :
-        function() { return this.impl[name]; };
+        new Function('return this.__impl4cf1e782hg__.' + name) :
+        function() { return this.__impl4cf1e782hg__[name]; };
   }
 
   function getSetter(name) {
     return hasEval && isIdentifierName(name) ?
-        new Function('v', 'this.impl.' + name + ' = v') :
-        function(v) { this.impl[name] = v; };
+        new Function('v', 'this.__impl4cf1e782hg__.' + name + ' = v') :
+        function(v) { this.__impl4cf1e782hg__[name] = v; };
   }
 
   function getMethod(name) {
     return hasEval && isIdentifierName(name) ?
-        new Function('return this.impl.' + name +
-                     '.apply(this.impl, arguments)') :
-        function() { return this.impl[name].apply(this.impl, arguments); };
+        new Function('return this.__impl4cf1e782hg__.' + name +
+                     '.apply(this.__impl4cf1e782hg__, arguments)') :
+        function() {
+          return this.__impl4cf1e782hg__[name].apply(
+              this.__impl4cf1e782hg__, arguments);
+        };
   }
 
   function getDescriptor(source, name) {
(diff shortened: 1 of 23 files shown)


arv:
@sorvell @sjmiles @azakus 

Heads up! I think you might have been using `.impl` before.

arv:
Ping

arv:
Actually, lets wait with this one a little. Let's land it (if you are OK with it) just after next release branch.

dfreedm:
@arv, What's the motivation for changing the wrapping strategy?
Seems like a pretty big change.

arv:
How is this a big change? It just renames `.impl`.

I want to potentially switch to not using wrappers and just replace the properties where it is supported.

Also, this change enables us to do cross frame DOM access if we want to support that.

arv:
How about just renaming it to `impl_`?

arv:
@sjmiles @sorvell 

Here is the change I was talking about.

As you can see `isWrapper` and `isNative` are now just checking for the presence of this property, allowing cross frame uwrapping.

Wrapping an object from a different frame would need some more work to find the correct wrapper. I have some ideas here too.

arv:
@jmesserly What do you think about this? I still think this is a good change.

src/wrappers.js
```diff
@@ -110,23 +110,31 @@ window.ShadowDOMPolyfill = {};
     return /^\w[a-zA-Z_0-9]*$/.test(name);
   }
 
+  // The name of the implementation property is intentionally hard to
+  // remember. Unfortunately, browsers are slower doing obj[expr] than
+  // obj.foo so we resort to repeat this ugly name. This ugly name is never
+  // used outside of this file though.q
```
jmesserly: extra "q" at the end of this comment

src/wrappers.js
```diff
@@ -110,23 +110,31 @@ window.ShadowDOMPolyfill = {};
...
+
   function getGetter(name) {
     return hasEval && isIdentifierName(name) ?
-        new Function('return this.impl.' + name) :
-        function() { return this.impl[name]; };
+        new Function('return this.__impl4cf1e782hg32d07__.' + name) :
```
jmesserly: really crazy would be to regenerate this per build, but that seems like serious overkill :)

src/wrappers.js
```diff
@@ -250,38 +258,12 @@ window.ShadowDOMPolyfill = {};
...
-
   function isWrapper(object) {
-    return object instanceof wrappers.EventTarget ||
-           object instanceof wrappers.Event ||
-           object instanceof wrappers.Range ||
-           object instanceof wrappers.DOMImplementation ||
-           object instanceof wrappers.CanvasRenderingContext2D ||
-           wrappers.WebGLRenderingContext &&
-               object instanceof wrappers.WebGLRenderingContext;
+    return object && object.__impl4cf1e782hg32d07__;
```
jmesserly: very nice!

jmesserly:
lgtm. Nice change.

arv:
@azakus Are you still concerned about this change?

dfreedm:
No, I think this is fine. Needs a rebaseline though.

arv:
Rebased and updated new code.

PTAL

jmesserly:
LGTM :)

nazar-pc:
My code not working after update to 0.4.0 and I think this pull request caused that.
For example, I've used `.impl` in this case:

``` javascript
player_element = document.createElement('audio');
context = new AudioContext;
source = context.createMediaElementSource(player_element.impl);
```

Because if I don't I get:

> TypeError: Argument 1 of AudioContext.createMediaElementSource does not implement interface HTMLMediaElement.

I have no idea how to fix this now. Can I rely on that weird `.__impl4cf1e782hg__` in future versions?

nazar-pc:
BTW, `does not implement interface HTMLMediaElement` - why not?

jmesserly:
You can use ShadowDOMPolyfill.unwrap(player_element) ... Needless to say,
don't use **impl***   :)

(edit: formatting)

nazar-pc:
I think this is too far from obvious and will cause huge compatibility issues when using with existing code.

arv:
@nazar-pc This was actually done to reduce the compatibility issues. Using wrap/unwrap allows us to do the right thing on all browsers.

PR #392, opened 2014-03-03
Wrap setDragImage

arv:
Since DataTransfer only has one method that requires unwrapping of a param we just override that method instead of wrapping the whole interface.

### DIFF
--- a/build.json
+++ b/build.json
@@ -41,5 +41,6 @@
   "src/wrappers/Selection.js",
   "src/wrappers/Document.js",
   "src/wrappers/Window.js",
+  "src/wrappers/DataTransfer.js",
   "src/wrappers/override-constructors.js"
 ]
--- a/shadowdom.js
+++ b/shadowdom.js
@@ -57,6 +57,7 @@
     'src/wrappers/Selection.js',
     'src/wrappers/Document.js',
     'src/wrappers/Window.js',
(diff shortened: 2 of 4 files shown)


arv:
@sorvell Any ideas how to unit test this?

dfreedm:
I'm not aware of any ways to unit test this. Drag and Drop cannot be triggered programmatically by MouseEvents, so you have to physically drag and drop to check.

I guess you could add a small page that loads shadowdom, then manually check every once in a while.

arv:
I added a manual test.

PTAL

arv:
Ping

sorvell:
Can we use an svg without the profanity in the license? Just seems unnecessary.

arv:
Fair enough. I'll switch to a simple data url image instead. Less risk

On Thu, Mar 6, 2014 at 10:59 AM, Steve Orvell notifications@github.comwrote:

> Can we use an svg without the profanity in the license? Just seems
> unnecessary.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/Polymer/ShadowDOM/pull/392#issuecomment-36902610
> .

## 

erik

arv:
Done. PTAL

sorvell:
LGTM, thx

PR #398, opened 2014-03-11
HTMLSelectElement remove()

arv:
Fixes #395


dfreedm:
_sigh_, we can never have all the nice things

LGTM

PR #401, opened 2014-03-12
Work around IE bug related to relatedTarget

arv:
We've  seen cases where the relatedTarget as reported by IE returns an Object without any properties and without an [[Prototype]].


dfreedm:
LGTM

PR #402, opened 2014-03-13
Propage TreeScope into the shadow trees too.

arv:
When updating the tree scope of a subtree we need to go into the shadow root subtree to update its parent.

Fixes #397


arv:
@sorvell Thanks for the help to reproduce this.

dfreedm:
LGTM!

PR #403, opened 2014-03-13
Fix issue related to returnValue in BeforeUnloadEvent

### LINKED ISSUE #445: addEventListener beforeunload not working
Hey, I am using a external library that uses setInterval and adds a beforeunload listener to clean it before leaving the page. But thats not happening, so it is causing the user to be prompted a message before leaving.

It seems that polymer is overriding the addEventListener and not firing the beforeunload properly. 

Is it a known issue? Is there any fix?

Thanks a lot.

arv:
Fixes https://github.com/Polymer/polymer/issues/445

### DIFF
--- a/src/wrappers/events.js
+++ b/src/wrappers/events.js
@@ -170,8 +170,17 @@
     if (handledEventsTable.get(originalEvent))
       return;
     handledEventsTable.set(originalEvent, true);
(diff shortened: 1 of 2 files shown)


dfreedm:
LGTM

PR #404, opened 2014-03-14
IE & relatedTarget again

arv:
PR #401 was not complete. My intention was to set the relatedTarget
to null if we run into one of these invalid relatedTargets.

### DIFF
--- a/src/wrappers/events.js
+++ b/src/wrappers/events.js
@@ -276,19 +276,20 @@
       // way to have relatedTarget return the adjusted target but worse is that
       // the originalEvent might not have a relatedTarget so we hit an assert
       // when we try to wrap it.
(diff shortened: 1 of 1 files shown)


src/wrappers/events.js
```diff
@@ -271,24 +271,26 @@
     if ('relatedTarget' in event) {
       var originalEvent = unwrap(event);
       var unwrappedRelatedTarget = originalEvent.relatedTarget;
+      var adjusted;
```
dfreedm: Is this needed outside the `unwrappedRelatedTarget` branch?

arv: Removed.

dfreedm:
LGTM

PR #410, opened 2014-03-28
Events: Fix issue where we ignored null relatedTargets

arv:
We set the relatedTarget to null in a few cases to work around bugs
in IE, but we ignored null due to the falsey check in the getter.


arv:
Ping

dfreedm:
LGTM

PR #411, opened 2014-03-31
Allow compareDocumentPosition to take a raw Node

arv:
jQueryUI uses `element.compareDocumentPosition(document.body)` so
we need to allow a non wrapper as the argument here.

Fixes #410


dfreedm:
Doesn't this actually fix #409?

dfreedm:
LGTM. Fixes #409

arv:
Yeah, this was #409 and not #410

PR #413, opened 2014-04-01
Rethrow erors in events

arv:
Instead of manually calling window.onerror, rethrow the first error
at the end of the dispatch. This way window.onerror gets all the
expected arguments which is important for error reporters.

Fixes #408


arv:
Ping

sorvell:
LGTM

PR #414, opened 2014-04-07
fixed hasEval in case of Chrome Apps

dinhvh:
Checking for document.securityPolicy.allowsEval won't work in that environment.

### DIFF
--- a/src/wrappers.js
+++ b/src/wrappers.js
@@ -16,6 +16,11 @@ window.ShadowDOMPolyfill = {};
   // even when the exception is caught
   var hasEval = !('securityPolicy' in document) ||
       document.securityPolicy.allowsEval;
+
+  if (typeof chrome !== 'undefined' && chrome.app && chrome.app.runtime) {
+    hasEval = false;
+  }
+
   if (hasEval) {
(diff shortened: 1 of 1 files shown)


src/wrappers.js
```diff
@@ -16,6 +16,11 @@ window.ShadowDOMPolyfill = {};
   // even when the exception is caught
   var hasEval = !('securityPolicy' in document) ||
       document.securityPolicy.allowsEval;
+
+  if (chrome && chrome.app && chrome.app.runtime) {
```
arv: This will not work in non chrome browsers. Use `typeof chrome !== 'undefined'` or `window.chrome`.

arv:
...but... doesn't Chrome App set the securityPolicy?

dinhvh:
It doesn't set the securityPolicy.

dfreedm:
Polymer/observe-js#44

dinhvh:
Updated. ptal. Thanks!

arv:
LGTM

Can you squash these into one commit?

dinhvh:
Done.

PR #415, opened 2014-04-08
Minor tweaks to reuse objects and skip forEach during boot

arv:
Reuse the descriptor objects where possible. For non generational
GC this reduces the GC churn.

Use c-style for loops instead of forEach since forEach is slow


arv:
Ping

dfreedm:
LGTM

PR #416, opened 2014-04-08
Make the window.onerror test more resilient to old browsers

arv:
Check the number of arguments passed to window.onerror before
asserting the value of the argument.


arv:
Ping

dfreedm:
LGTM

PR #418, opened 2014-04-09
Make sure we use the wrapper NodeList in querySelector.js

arv:


dfreedm:
How did this ever work before? Maybe because `window.NodeList` was already overridden by this point?

Either way, LGTM.

arv:
window.NodeList gets overridden but I've gotten reports that Safari still uses the builtin NodeList in some edge cases.

PR #419, opened 2014-04-09
Fix getElementsByTagName to handle colon and more

### LINKED ISSUE #468: Including platform.js breaks YouTube's iframe API
I've been playing around with Polymer for a few days now and it's awesome, thanks! Apologies if this is not an issue with Polymer, but rather YouTube's API.

---

Including platform.js on a page that is making use of YouTube's iframe API causes the following: **Uncaught SyntaxError: Failed to execute 'webkitMatchesSelector' on 'Element': 'yt:player' is not a valid selector.**

Here is a test case the reproduces the issue (removing the inclusion of platform.js on line 4 fixes the problem):

``` html
<!doctype html>
<html>
    <head>
        <script src="http://cdnjs.cloudflare.com/ajax/libs/polymer/0.2.2/platform.js"></script>
    </head>

    <body>
        <div id="ytplayer"></div>
        <script>
            // Load the IFrame Player API code asynchronously.
            var tag = document.createElement('script');
            tag.src = "https://www.youtube.com/player_api";
            var firstScriptTag = document.getElementsByTagName('script')[0];
            firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);

            // Replace the 'ytplayer' element with an <iframe> and
            // YouTube player after the API code downloads.
            var player;
            function onYouTubePlayerAPIReady() {
                player = new YT.Player('ytplayer', {
                    height: '390',
                    width: '640',
                    videoId: 'M7lc1UVf-VE'
                });
            }
        </script>
    </body>
</html>

```

arv:
Instead of relying on CSS selectors, this now implements the
matching as described by dom.spec.whatwg.org

Fixes https://github.com/Polymer/polymer/issues/468

### DIFF
--- a/src/querySelector.js
+++ b/src/querySelector.js
@@ -5,6 +5,7 @@
 (function(scope) {
   'use strict';
 
+  var HTMLCollection = scope.wrappers.HTMLCollection;
   var NodeList = scope.wrappers.NodeList;
 
   function findOne(node, selector) {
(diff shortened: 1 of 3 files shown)


arv:
@azakus One more for you

arv:
rebased

dfreedm:
Thx, LGTM

PR #423, opened 2014-04-22
Bring shadow dom algorithms up to date

arv:
This is a complete update of the distribution, composition and event
path algorithms from the spec draft.

Fixes #420

### DIFF
--- a/src/ShadowRenderer.js
+++ b/src/ShadowRenderer.js
@@ -97,29 +97,18 @@
     parentNode.removeChild(node);
   }
 
-  var distributedChildNodesTable = new WeakMap();
-  var eventParentsTable = new WeakMap();
-  var insertionParentTable = new WeakMap();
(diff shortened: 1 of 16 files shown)


arv:
@sjmiles @sorvell 

I know this is big but what is your time frame for reviewing this?

dfreedm:
I'll take a look

dfreedm:
Looks like something in here is crashing IE 11 tabs.

arv:
You know what. I saw that too. But then I checked out a really old version and it still crashed. So, after resetting IE's cache it no longer crashed on head. Maybe I need to spend some more time on figuring out why it crashed.

arv:
And now I cannot get IE11 to crash any more.

arv:
PTAL

dfreedm:
LGTM

PR #427, opened 2014-04-28
Implement TouchEvent wrappers

arv:
Fixes #425


dfreedm:
LGTM. There is a small risk of incompatibilities due to Safari not implementing all the aspects of the spec (TouchList.item), but that's probably pretty negligible.

arv:
Google Maps uses TouchList.prototype.item so I had to support that.

PR #428, opened 2014-04-28
Remove document.securityPolicy since it is no longer supported

arv:


arv:
Same fix as for https://github.com/Polymer/observe-js/issues/50 but for ShadowDOM.

arv:
@azakus

dfreedm:
LGTM

PR #439, opened 2014-05-20
appendChild from a tree with a rendered causes failures

arv:
The following fails to correctly update the internal pointers.

treeWithoutRenderer.appendChild(childFromParentWithRenderer)

Fixes #438


sorvell:
LGTM thx for the fix

PR #440, opened 2014-05-21
Listeners added in the current dispatch should not get called

### LINKED ISSUE #511: Adding an event handler in an event handler can lead to infinite looping
Example here:

https://gist.github.com/jcmoore/38135abefad78dc41e79#file-examlple-html-L14-L30

Not sure what the implications would be, but a solution might be as simple as using the "initial" length of the listeners array in the loop condition (instead of using the "current" length) for ShadowDOM/src/wrappers/events.js' invoke() method:

https://gist.github.com/jcmoore/38135abefad78dc41e79#file-shadowdom-src-wrappers-events-js-L385-L387

arv:
Fixes https://github.com/Polymer/polymer/issues/511

### DIFF
--- a/src/wrappers/events.js
+++ b/src/wrappers/events.js
@@ -382,7 +382,7 @@
     targetTable.set(event, target);
     currentTargetTable.set(event, currentTarget);
 
-    for (var i = 0; i < listeners.length; i++) {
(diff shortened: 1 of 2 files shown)


dfreedm:
LGTM

PR #447, opened 2014-05-29
Event path should return every nodes in the event path

arv:
See https://github.com/w3c/webcomponents/commit/2a44fb1a32a3a6aaa7bd82d37f6bd17eb2d67535 for spec change.


dfreedm:
LGTM!

PR #448, opened 2014-05-30
Fix issue with DOMTokenList

arv:
The build.json file did not include DOMTokenList leading to
classList being brokon in 0.3.0

This adds a test that ensures that the list of files is in sync so
that these kind of errors do not happen again.

Fixes #445

### DIFF
--- a/build.json
+++ b/build.json
@@ -14,6 +14,7 @@
   "src/wrappers/node-interfaces.js",
   "src/wrappers/CharacterData.js",
   "src/wrappers/Text.js",
+  "src/wrappers/DOMTokenList.js",
   "src/wrappers/Element.js",
   "src/wrappers/HTMLElement.js",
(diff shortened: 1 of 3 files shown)


arv:
@azakus @sjmiles @sorvell 

This is a pretty severe bug. It caused 0.3.0 to not work when using the compiled platform.js.

sorvell:
Thx Arv, we'll prioritize doing another release.

dfreedm:
Sad we didn't catch this in release testing :(

LGTM

arv:
This caused build failures:

http://build.chromium.org/p/client.polymer/builders/ShadowDOM%20Mac/builds/142

@morethanreal Is the build.json file accessible by the bots?

dfreedm:
Karma won't serve it automatically

arv:
https://github.com/Polymer/ShadowDOM/pull/449

PR #449, opened 2014-05-30
Karma needs to serve build.json since the test depends on that

arv:


dfreedm:
Closing in favor of #450

PR #453, opened 2014-06-06
Scope elementFromPoint to the current TreeScope

dfreedm:
Fixes #452

### DIFF
--- a/src/wrappers/events.js
+++ b/src/wrappers/events.js
@@ -815,6 +815,15 @@
     if (!element)
       return null;
     var path = getEventPath(element, null);
+
+    // scope the path to this TreeScope
+    var idx = path.lastIndexOf(self);
+    if (idx == -1)
+      return null;
+    else
+      path = path.slice(0, idx);
+
+    // TODO(dfreedm): pass idx to eventRetargetting to avoid array copy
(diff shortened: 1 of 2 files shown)


src/wrappers/events.js
```diff
@@ -815,6 +815,14 @@
     if (!element)
       return null;
     var path = getEventPath(element, null);
+
+    // scope the path to this TreeScope
+    var idx = path.indexOf(self);
```
arv: lastIndexOf seems conceptually more correct since searching from the leaf towards the root seems more logical.

src/wrappers/events.js
```diff
@@ -815,6 +815,14 @@
...
+    if (idx == -1)
+      return null;
+    else
+      path = path.slice(0, idx);
```
arv, line 824: Maybe we could change eventRetargetting to take the index to start at so we don't have to copy the array.

I don't think the slice is a big deal though. Depends on how hot this code is.

arv:
LGTM

dfreedm:
Using lastIndexOf, added a TODO for looking at passing index to eventRetargetting if we think it becomes a hotpath.

arv:
LGTM2

PR #455, opened 2014-06-09
SVGElement does not have classList in IE

arv:
We therefore move it to HTMLElement so that code that does feature
detection of classList will work correctly in IE.

Fixes #454

### DIFF
--- a/src/wrappers/SVGElement.js
+++ b/src/wrappers/SVGElement.js
@@ -5,12 +5,24 @@
 (function(scope) {
   'use strict';
 
+  var Element = scope.wrappers.Element;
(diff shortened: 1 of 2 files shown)


dfreedm:
LGTM

arv:
This broke the build. Investigating.

dfreedm:
Looks like `svg_el.classList instanceof DOMTokenList` broke on Safari and Chrome.

arv:
Found the issue. New PR coming.

arv:
https://github.com/Polymer/ShadowDOM/pull/456

PR #457, opened 2014-06-10
Fix issue with event reentrancy

arv:
If an inner invoke removes a listener, the outer invoke now loops
over the stale listener length. To solve this we only clean up the
listener array when we are in the outermost invoke.


dfreedm:
LGTM

PR #461, opened 2014-06-26
Fix issue with invalidation

arv:
When we invalidate a renderer we also need to invalidate the parent
renderer. This is important because nodes from the parent renderer
might need to be reprojected.

Fixes #460


arv:
@morethanreal @azakus

sorvell:
LGTM

PR #462, opened 2014-06-26
fix form.elements -- this is used by NodeBind getAssociatedRadioButtons

jmesserly:
form.elements should return an HTMLCollection.

this code in [NodeBind.js](https://github.com/Polymer/NodeBind/blob/master/src/NodeBind.js#L208)

``` js
     return filter(element.form.elements, function(el) {
        return el != element &&
            el.tagName == 'INPUT' &&
            el.type == 'radio' &&
            el.name == element.name;
      });
```

...uses form.elements, which was returning unwrapped nodes under SD polyfill.  This later causes an error because Node.bind stores bindings_ on the wrapper, and when it doesn't find the data it stored, blows up with a TypeError.

Caught this by running NodeBind tests with platform.js (normally they don't run with SD polyfill -- maybe they should?)

### DIFF
--- a/src/wrappers/HTMLFormElement.js
+++ b/src/wrappers/HTMLFormElement.js
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2014 The Polymer Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file.
+ */
+
+(function(scope) {
+  'use strict';
+
+  var HTMLElement = scope.wrappers.HTMLElement;
+  var mixin = scope.mixin;
+  var registerWrapper = scope.registerWrapper;
+  var wrapHTMLCollection = scope.wrapHTMLCollection;
+  var unwrap = scope.unwrap;
+
+  var OriginalHTMLFormElement = window.HTMLFormElement;
+
+  function HTMLFormElement(node) {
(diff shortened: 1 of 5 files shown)


jmesserly:
Hey, @arv do you mind taking a look?

I patterned it after similar wrappers like HTMLTableRowElement. All tests pass after this change.

src/wrappers/HTMLFormElement.js
```diff
@@ -0,0 +1,36 @@
...
+
+(function(scope) {
+  'use strict';
+
+  var HTMLElement = scope.wrappers.HTMLElement;
+  var mixin = scope.mixin;
+  var registerWrapper = scope.registerWrapper;
+  var wrapHTMLCollection = scope.wrapHTMLCollection;
+  var unwrap = scope.unwrap;
+  var wrap = scope.wrap;
```
arv: It doesn't look like you are using `wrap`.

arv:
LGTM

HTMLFormControlsCollection is one of the worst abominations of the DOM APIs. It has named interceptors too and sometimes these return an element and sometime another collection. Hopefully we do not have to ever implement that.

src/wrappers/HTMLFormElement.js
jmesserly: good catch. fixed!

PR #463, opened 2014-07-01
Add FormData wrapper

### LINKED ISSUE #587: FormData.constructor fails when passed HTMLFormElement in Firefox
I ran a very simple test in Firefox 30 OSX: I attempted to pass an `HTMLFormElement` into a `FormData` constructor.  In the absence of platform.js, the attempt succeeds.  As soon as I import platform.js, it fails.

Please see [this fiddle](http://jsfiddle.net/S6LH3/) for a one-line demo of the issue.  When run in Firefox as is, with the console open, you'll see the error.  If you remove the platform.js import, it executes without error.

Is Polymer either messing with the `FormData` constructor, or perhaps `HTMLFormElement`?

arv:
Fixes https://github.com/Polymer/polymer/issues/587

### DIFF
--- a/build.json
+++ b/build.json
@@ -45,5 +45,6 @@
   "src/wrappers/Document.js",
   "src/wrappers/Window.js",
   "src/wrappers/DataTransfer.js",
+  "src/wrappers/FormData.js",
   "src/wrappers/override-constructors.js"
 ]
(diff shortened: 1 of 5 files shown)


dfreedm:
LGTM

PR #472, opened 2014-07-21
Add wrap examples

robdodson:
Per https://github.com/Polymer/platform/issues/69
This adds some examples to demonstrate how and when to use wrap

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,10 @@ There are bound to be cases where we haven't done the wrapping for you. In those
 
 ex:
 
-    wrap(document).addEventListener(...)
+    wrap(document.body)
+    // or get body of the wrapped document
+    wrap(document).body
+
     unwrap(div).firstChild instanceof HTMLElement
 
 If you plan to work with elements that need to be wrapped over and over, try passing a wrapped version of the element into an immediately-invoked function expression.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,6 +88,23 @@ The intent is to wrap all the DOM objects that interact with the DOM tree. For t
 
 There are bound to be cases where we haven't done the wrapping for you. In those cases you can use `wrap` to create a wrapper of a native object, or `unwrap` to get the underlying native object from a wrapper. These two functions are available on the `ShadowDOMPolyfill` object.
 
+ex:
+
+    wrap(document).addEventListener(...)
```
arv: This is a bad example since we can wrap methods of `window` and `document` and we have wrapped this one, and the most common ones already. The problematic ones are properties of window and document. Most common issue is `document.body` but also `document.documentElement`

``` js
wrap(document.body)
// or get body of the wrapped document
wrap(document).body
wrap(document).head
wrap(document).documentElement
wrap(document).firstChild
```

arv:
Thanks

robdodson:
OK, I updated it to use document.body

arv:
LGTM

PR #473, opened 2014-07-22
Disable eval for Firefox OS Apps

arv:
Fixes #467


jmesserly:
nice, lgtm. Saw you land same fix land in observe-js :)

PR #474, opened 2014-07-25
Implement newly proposed DOM selection approach

jcmoore:
There may be some possible improvements for selecting light-DOM nodes in a shadow-rooted element with no contained content element -- I also tried querying the whole DOM natively and traversing it instead of the wrapped tree (https://gist.github.com/jcmoore/69a49f887e03c1d2bfa1#file-queryselector2-js) but it was no faster than the original method, and probably had significant memory implications.

### DIFF
--- a/src/querySelector.js
+++ b/src/querySelector.js
@@ -7,6 +7,39 @@
 
   var HTMLCollection = scope.wrappers.HTMLCollection;
   var NodeList = scope.wrappers.NodeList;
+  var getTreeScope = scope.getTreeScope;
+  var wrap = scope.wrap;
+
+  var originalDocumentQuerySelector = document.querySelector;
+  var originalElementQuerySelector = document.documentElement.querySelector;
+
+  var originalDocumentQuerySelectorAll = document.querySelectorAll;
+  var originalElementQuerySelectorAll = document.documentElement.querySelectorAll;
+
+  var originalDocumentGetElementsByTagName = document.getElementsByTagName;
+  var originalElementGetElementsByTagName = document.documentElement.getElementsByTagName;
+
+  var originalDocumentGetElementsByTagNameNS = document.getElementsByTagNameNS;
+  var originalElementGetElementsByTagNameNS = document.documentElement.getElementsByTagNameNS;
+
+  var OriginalElement = window.Element;
+  var OriginalDocument = window.HTMLDocument;
+
+  function filterNodeList(list, index, result) {
+    var wrappedItem = null;
+    var root = null;
+    for (var i = 0, length = list.length; i < length; i++) {
+      wrappedItem = wrap(list[i]);
+      if (root = getTreeScope(wrappedItem).root) {
+        if (root instanceof scope.wrappers.ShadowRoot) {
+          continue;
+        }
+      }
+      result[index++] = wrappedItem;
+    }
+    
+    return index;
+  }
 
   function findOne(node, selector) {
     var m, el = node.firstElementChild;
@@ -37,7 +70,7 @@
     return true;
   }
 
-  function matchesLocalName(el, localName) {
+  function matchesLocalNameOnly(el, ns, localName) {
     return el.localName === localName;
   }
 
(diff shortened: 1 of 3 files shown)


jcmoore:
https://github.com/Polymer/polymer/issues/629

src/querySelector.js
```diff
@@ -7,6 +7,41 @@
 
   var HTMLCollection = scope.wrappers.HTMLCollection;
   var NodeList = scope.wrappers.NodeList;
+  var wrap = scope.wrapIfNeeded;
```
arv: keep this as wrapIfNeeded to avoid confusion

src/querySelector.js
```diff
@@ -7,6 +7,41 @@
...
+
+  var documentQuerySelector = document.querySelector;
```
arv: we usually name these original\* to make it clearer

src/querySelector.js
```diff
@@ -7,6 +7,41 @@
...
+  var OriginalDocument = window.HTMLDocument;
+  var OriginalFragment = window.DocumentFragment;
+
+  function filterNodeList(list, recur, p, index, result, arg0, arg1) {
+    if (list == null)
+      return list;
+    var wrappedItem = null;
+    var root = null;
+    for (var i = 0, length = list.length; i < length; i++) {
+      wrappedItem = wrap(list[i]);
```
arv, line 32: this should be wrap and not wrapIfNeeded. The list must not contain any wrapped elements at this point.

src/querySelector.js
```diff
@@ -7,6 +7,41 @@
...
+      if (root = wrappedItem.treeScope_ && wrappedItem.treeScope_.root) {
```
arv: Use `getTreeScope(wrappedItem)`

src/querySelector.js
```diff
@@ -7,6 +7,41 @@
...
+        if (root.impl instanceof OriginalFragment) {
```
arv: root instance of scope.wrappers.DocumentFragment

arv: Or maybe instanceof ShadowRoot? It is not clear to me what the intent is here?

src/querySelector.js
```diff
@@ -7,6 +7,41 @@
...
+  var elementGetElementsByTagName = document.documentElement.getElementsByTagName;
+
+  var documentGetElementsByTagNameNS = document.getElementsByTagNameNS;
+  var elementGetElementsByTagNameNS = document.documentElement.getElementsByTagNameNS;
+
+  var OriginalElement = window.Element;
+  var OriginalDocument = window.HTMLDocument;
+  var OriginalFragment = window.DocumentFragment;
+
+  function filterNodeList(list, recur, p, index, result, arg0, arg1) {
```
arv: Remove unused parameters

src/querySelector.js
```diff
@@ -49,40 +84,118 @@
...
+      return filterNodeList(list, querySelectorAllFiltered, p,
+          index,
+          result,
+          selector,
+          null);
+  }
+
   var SelectorsInterface = {
     querySelector: function(selector) {
+      var target = this.impl || this;
```
arv: I don't like that we are mixing and matching wrapped nodes and non wrapped nodes. We should stick with one and have the caller wrap/unwrap as needed.

src/querySelector.js
```diff
@@ -49,40 +84,118 @@
...
+          0,
+          result,
+          selector);
+
+      return result;
     }
   };
 
+  function getElementsByTagNameFiltered (p, index, result, localName, lowercase) {
+      var target = this.impl || this;
```
arv: wrong indentation

src/querySelector.js
```diff
@@ -49,40 +84,118 @@
...
+          ns,
+          localName);
+  }
+
   var GetElementsByInterface = {
     getElementsByTagName: function(localName) {
       var result = new HTMLCollection();
-      if (localName === '*')
-        return findElements(this, result, matchesEveryThing);
+      var match = (localName === '*') ? matchesEveryThing : matchesTagName;
```
arv: useless parens

test/js/Element.js
```diff
@@ -15,6 +15,7 @@ suite('Element', function() {
     div = null;
   });
 
+  if (0) // this test fails due to patch of querySelector.js (issue #629)
```
arv: or skipTest or skip

function skipTest() {}

replace test with skipTest

src/querySelector.js
jcmoore: So my intent is to determine if a node is in the shadow-DOM.  I see now that ShadowRoot is the type the wrapped instances I'm looking for.  Thanks

src/querySelector.js
```diff
@@ -7,6 +7,37 @@
...
+  var OriginalElement = window.Element;
+  var OriginalDocument = window.HTMLDocument;
+
+  function filterNodeList(list, index, result) {
+    var wrappedItem = null;
+    var root = null;
+    for (var i = 0, length = list.length; i < length; i++) {
+      wrappedItem = wrap(list[i]);
+      if (root = getTreeScope(wrappedItem).root)
+        if (root instanceof ShadowRoot)
```
jcmoore: I'm relying on the global/window reference to ShadowRoot here, because it is not on "scope.wrappers" when this module loads...

Should I instead refer to the (un-cached) reference directly from "scope" and/or where can I change the loading order of the modules (so ShadowRoot goes first)?

arv: > Should I instead refer to the (un-cached) reference directly from "scope"

Yup. We do this in a few other places where we get these dependency cycles.

src/querySelector.js
```diff
@@ -49,40 +80,99 @@
...
+    return index;
   }
 
   // find and findAll will only match Simple Selectors,
   // Structural Pseudo Classes are not guarenteed to be correct
   // http://www.w3.org/TR/css3-selectors/#simple-selectors
 
+  function querySelectorAllFiltered (p, index, result, selector) {
+    var target = this.impl;
+    var list = null;
```
arv: `var list;` is sufficient here

src/querySelector.js
```diff
@@ -49,40 +80,99 @@
...
+    if (target instanceof OriginalElement)
+      list = originalElementQuerySelectorAll.call(target, selector);
+    else if (target instanceof OriginalDocument)
+      list = originalDocumentQuerySelectorAll.call(target, selector);
+
+    if (!list)
```
arv: When does this happen? Maybe add a comment when the else case is reached?

src/querySelector.js
arv: Same in a few more places

src/querySelector.js
jcmoore: As far as I know, it only happens on ShadowRoot instances (which have DocumentFragments in them).  I was trying to be safe and just fallback to the original traversal in the more "complex" cases.

I'll add a comment

ebidel:
Before this PR can be merged, please sign the CLA. See the [CONTRIBUTING](http://polymer.github.io/CONTRIBUTORS.txt) guide.

jcmoore:
I'd already received confirmation that the MediaMath's (corporate) CLA had been accepted (cla-submissions@google.com), do I need to send a copy somewhere else?

ebidel:
Got it. Thanks for the heads up.

ebidel:
Thanks for signing the CLA!

src/querySelector.js
arv: I would prefer if this was scope.wrappers.ShadowRoot so that we do not have any free variables in here.

src/querySelector.js
```diff
@@ -49,40 +80,99 @@
...
 
+  function querySelectorAllFiltered (p, index, result, selector) {
+    var target = this.impl;
+    var list;
+    if (target instanceof OriginalElement)
+      list = originalElementQuerySelectorAll.call(target, selector);
+    else if (target instanceof OriginalDocument)
+      list = originalDocumentQuerySelectorAll.call(target, selector);
+
+    if (!list) // occurs when when this/target -> ShadowRoot/DocumentFragment
```
arv: How about?

// When we get a ShadowRoot the logical tree is going to be disconnected so we do a manual tree traversal

Any reason this is not just the else branch?

``` js
if (target instanceof OriginalElement)
  list = originalElementQuerySelectorAll.call(target, selector);
else if (target instanceof OriginalDocument)
  list = originalDocumentQuerySelectorAll.call(target, selector);
else
  return findElements(this, index, result, p, selector, null);
```

src/querySelector.js
```diff
@@ -49,40 +80,99 @@
...
+      return findElements(this, index, result, p, selector, null);
+
+    return filterNodeList(list, index, result);
+  }
+
   var SelectorsInterface = {
     querySelector: function(selector) {
+      var target = this.impl;
+      if (target instanceof OriginalElement)
+        return wrap(originalElementQuerySelector.call(target, selector));
```
arv: We should filter this one too

src/querySelector.js
jcmoore: Ah, I misunderstood your "Yup".

src/querySelector.js
```diff
@@ -49,40 +80,114 @@
...
   // http://www.w3.org/TR/css3-selectors/#simple-selectors
 
+  function querySelectorAllFiltered (p, index, result, selector) {
+    var target = this.impl;
+    var list;
+    if (target instanceof OriginalElement)
+      list = originalElementQuerySelectorAll.call(target, selector);
+    else if (target instanceof OriginalDocument)
+      list = originalDocumentQuerySelectorAll.call(target, selector);
+    else
```
arv: These now need curlies {}

``` js
    if (target instanceof OriginalElement) {
      list = originalElementQuerySelectorAll.call(target, selector);
    } else if (target instanceof OriginalDocument) {
      list = originalDocumentQuerySelectorAll.call(target, selector);
    } else {
      // When we get a ShadowRoot the logical tree is going to be disconnected
      // so we do a manual tree traversal
      return findElements(this, index, result, p, selector, null);
    }
```

src/querySelector.js
```diff
@@ -49,40 +80,114 @@
...
     querySelector: function(selector) {
-      return findOne(this, selector);
+      var target = this.impl;
+      var wrappedItem;
+      var root;
+      if (target instanceof OriginalElement)
+        wrappedItem = wrap(originalElementQuerySelector.call(target, selector));
+      else if (target instanceof OriginalDocument)
+        wrappedItem = wrap(originalDocumentQuerySelector.call(target, selector));
+      else
```
arv: same here

src/querySelector.js
```diff
@@ -7,6 +7,37 @@
...
+
+  var OriginalElement = window.Element;
+  var OriginalDocument = window.HTMLDocument;
+
+  function filterNodeList(list, index, result) {
+    var wrappedItem = null;
+    var root = null;
+    for (var i = 0, length = list.length; i < length; i++) {
+      wrappedItem = wrap(list[i]);
+      if (root = getTreeScope(wrappedItem).root)
```
arv: Our style guide requires curly braces when the statement does not fit on one line.

arv:
LGTM once you fix the curlies for the if/else. (You can use curlies for all if/else if you want too).

Can you squash these into a single commit?

We usually include "Fixes #N" in the commit message too.

src/querySelector.js
jcmoore: (For future reference,) can curly braces be included when the statement DOES fit on one line?

arv: Yes.

arv: Here is some history:
- The Google JavaScript style  guide always requires curlies.
- Chrome and Blink style guides only requires them if the statement does not fit on one line.

I'm tempted to go back to the Google style guide and not use the Chrome one...

jcmoore:
Ah, to do a squash on the commits do I need a new pull-request (on a new branch?

(I'm unfamiliar with the process I'm afraid)

arv:
No need for a new PR or a new branch:

http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request

jcmoore:
All squashed

arv:
Thank you so much. Lets hope this one does not cause any real world issues.

@sjmiles @sorvell Keep your eyes open.

sjmiles:
Thanks you guys. :) We'll check it out.

frankiefu:
Filed an issue seems to be related to this change. https://github.com/Polymer/ShadowDOM/issues/476

jcmoore:
(Very new to pull-request work-flow).  In order to address this, must I do a brand new pull-request?

arv:
Yeah. A new pull request seems appropriate.

PR #477, opened 2014-07-28
Handle native querySelector returning nothing

jcmoore:
```
- Fixes #476
```

### DIFF
--- a/src/querySelector.js
+++ b/src/querySelector.js
@@ -128,7 +128,11 @@
         return findOne(this, selector);
       }
 
-      if (root = getTreeScope(wrappedItem).root) {
+      if (!wrappedItem) {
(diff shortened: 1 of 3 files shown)


src/querySelector.js
```diff
@@ -128,7 +128,11 @@
         return findOne(this, selector);
       }
 
-      if (root = getTreeScope(wrappedItem).root) {
+      if (!wrappedItem) {
+          // When the original query returns nothing
```
arv: wrong indentation

arv:
LGTM. Thanks for fixing.

Would you mind fixing the style nit and squash?

jcmoore:
Should be all set now BTW,  thanks!

arv:
Thanks again.

PR #481, opened 2014-07-31
Account for DOM selection in sub-shadow elements

jcmoore:
```
- Fixes #480
```

### DIFF
--- a/test/js/Element.js
+++ b/test/js/Element.js
@@ -285,4 +285,58 @@ suite('Element', function() {
     assert.equal(as[0], a1);
     assert.equal(as.item(0), a1);
   });
+
+  test('sub shadow-root traversal', function() {
+    var div = document.createElement("DIV");
+    var sr = div.createShadowRoot();
+    sr.innerHTML = "<aa><bb></bb></aa>";
+
+    var saal = sr.getElementsByTagName("aa");
+    var sbbl = sr.getElementsByTagName("bb");
+    assert.equal(saal.length, 1);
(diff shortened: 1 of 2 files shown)


test/js/Element.js
```diff
@@ -285,4 +285,58 @@ suite('Element', function() {
...
+    var div = document.createElement("DIV");
+    var sr = div.createShadowRoot();
+    sr.innerHTML = "<aa><bb></bb></aa>";
+
+    var saal = sr.getElementsByTagName("aa");
+    var sbbl = sr.getElementsByTagName("bb");
+    assert.equal(saal.length, 1);
+    assert.equal(sbbl.length, 1);
+
+    var saa = saal [0];
```
arv, line 299: no ws before [

test/js/Element.js
```diff
@@ -285,4 +285,58 @@ suite('Element', function() {
     assert.equal(as[0], a1);
     assert.equal(as.item(0), a1);
   });
+
+  test('sub shadow-root traversal', function() {
+    var div = document.createElement("DIV");
```
arv, line 290: Use single quotes

Use 'div' (lower case)

arv:
LGTM. I'll fix the nits and land.

If this is not sufficient another option would be to only use the filtering when called as document.\* and let all Element and ShadowRoot use the old manual tree traversal.

arv:
Committed as 541b3524aa161e4f9b6d9cba231520b09f0f40ed

PR #484, opened 2014-08-04
unwrap(FormData) fails on Firefox

atotic:
unwrap(FormData) did not unwrap on Firefox. Traced it down to failure of isWrapper function. One line fix above.
I needed to unwrap FormData to pass it to xhr.send()

This fixes one case only. Looking at wrappers, looks like all of these are wrapped on FF, and only a small subset is valid in 'isWrapper'. This might be a bigger fix. I am not familiar with your internals, but it looks like isWrapper could be implemented as:
        if ('impl' in object)
testing instanceof walks prototype chain, and doing instanceof for all the wrappers below would be very slow.

wrappers in Firefox:

Audio
BeforeUnloadEvent
CanvasRenderingContext2D
CharacterData
Comment
CustomEvent
DOMImplementation
DOMTokenList
Document
DocumentFragment
Element
Event
EventTarget
FocusEvent
FormData
HTMLAudioElement
HTMLButtonElement
HTMLCanvasElement
HTMLCollection
HTMLContentElement
HTMLElement
HTMLFieldSetElement
HTMLFormElement
HTMLImageElement
HTMLInputElement
HTMLLabelElement
HTMLLegendElement
HTMLMediaElement
HTMLObjectElement
HTMLOptionElement
HTMLOutputElement
HTMLSelectElement
HTMLShadowElement
HTMLTableElement
HTMLTableRowElement
HTMLTableSectionElement
HTMLTemplateElement
HTMLTextAreaElement
HTMLUnknownElement
Image
MouseEvent
MutationObserver
Node
NodeList
Option
Range
SVGElement
SVGUseElement
Selection
ShadowRoot
Text
UIEvent
WebGLRenderingContext
Window

### DIFF
--- a/src/wrappers.js
+++ b/src/wrappers.js
@@ -301,6 +301,7 @@ window.ShadowDOMPolyfill = {};
            object instanceof OriginalRange ||
            object instanceof OriginalDOMImplementation ||
            object instanceof OriginalCanvasRenderingContext2D ||
+           object instanceof OriginalFormData ||
            OriginalWebGLRenderingContext &&
                object instanceof OriginalWebGLRenderingContext ||
            OriginalSVGElementInstance &&
--- a/src/wrappers.js
+++ b/src/wrappers.js
@@ -289,6 +289,7 @@ window.ShadowDOMPolyfill = {};
            object instanceof wrappers.Range ||
            object instanceof wrappers.DOMImplementation ||
            object instanceof wrappers.CanvasRenderingContext2D ||
+           object instanceof wrappers.FormData ||
            wrappers.WebGLRenderingContext &&
                object instanceof wrappers.WebGLRenderingContext;
   }
--- a/src/wrappers.js
+++ b/src/wrappers.js
@@ -282,7 +282,8 @@ window.ShadowDOMPolyfill = {};
   var OriginalCanvasRenderingContext2D = window.CanvasRenderingContext2D;
   var OriginalWebGLRenderingContext = window.WebGLRenderingContext;
   var OriginalSVGElementInstance = window.SVGElementInstance;
-
+  var OriginalFormData = window.FormData;
+  
   function isWrapper(object) {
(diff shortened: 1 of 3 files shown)


arv:
> if ('impl' in object)

Take a look at https://github.com/Polymer/ShadowDOM/pull/391/files#diff-a54a07abc94b9272d561c1f34dd9bac3R262

arv:
Can you update `isNative` too? These are supposed to compliment each other.

arv:
Can this be tested?

atotic:
Re: if ('impl' in object)
your patch looks like the the right way to fix this.

I've extended the patch for isNative.

Simple pseudocode test case would be:
var hasFormDataWrapper = isFirefox;
var fd = new FormData();
var fdu  = unwrap(fd)
assert(!hasFormDataWrapper || fd != fdu);

arv:
I meant, can you add a unit test? Every change requires a test.

Adding the following to test/js/FormData.js should be sufficient:

``` js
test('wrap/unwrap', function() {
  var fd = new FormData();
  var unwrapped = unwrap(fd);
  var wrapped = wrap(unwrapped);
  assert.equal(fd, wrapped);
});
```

You also need to add `var wrap = ShadowDOMPolyfill.wrap;` etc inside the suite.

atotic:
That was a bit uglier than expected, and passing the test required more bug fixing:
- unwrap/wrap combo would fail to wrap.
- because wrap() calls FormData(formDataImpl),  instead of FormData(formElement)
- fix is to make FormData constructor deal with getting formDataImpl correctly. Not clean, but it works.

Any other wrappers that do not expect impl as first argument might suffer from the same problem.

++ for testing!

arv:
LGTM

arv:
Committed as c4da63735ba6c00a7d6af5c1b118f84bd6a2e114

PR #487, opened 2014-08-10
Added IE 10 support

kaapa-s:
I use Polymer in my project, but there was a problem, that it was not working on IE 10. I figured out that window.HTMLDocument is not available in IE 10, there is window.Document object instead. 

After this change everything is working.


arv:
LGTM

PR #490, opened 2014-08-13
Bypassing element wrapping for elements that do not exist

jsoverson:
This simply shorts out of the wrapping process if the native element doesn't exist, which can be the case for media related elements (HTMLAudioElement, HTMLVideoElement, etc) in headless browsers or browsers explicitly compiled without libraries that enable support for such elements (not uncommon for custom CEF builds).

This is the most obvious remaining issue before phantomjs v2 has basic testable support for platform.js and polymer.

I submitted the CLA and no tests seem to be affected (in Chrome, Chrome Canary, Firefox, Safari). If anything's not in order, feel free to close the PR and apply the change yourself, it's a trivial one.

### DIFF
--- a/src/wrappers/HTMLAudioElement.js
+++ b/src/wrappers/HTMLAudioElement.js
@@ -12,6 +12,8 @@
 
   var OriginalHTMLAudioElement = window.HTMLAudioElement;
 
+  if (!OriginalHTMLAudioElement) return;
+
   function HTMLAudioElement(node) {
     HTMLMediaElement.call(this, node);
   }
--- a/src/wrappers/HTMLMediaElement.js
+++ b/src/wrappers/HTMLMediaElement.js
@@ -10,6 +10,8 @@
 
   var OriginalHTMLMediaElement = window.HTMLMediaElement;
 
+  if (!OriginalHTMLMediaElement) return;
+
   function HTMLMediaElement(node) {
     HTMLElement.call(this, node);
   }


arv:
I'm not sure this is the right solution. For example, if there is no HTMLAudioElement we still set up the global HTMLAudioElement. I think we need  to handle this on a case by case basis.

jsoverson:
Probably, but the protection against undefined values still is still relevant and, given this is an edge case, worrying about it may not be worth the effort. This quick check gets the polyfill working beyond where it does right now.

arv:
The problem is that this stop gap will lead to other bugs which might be harder to track down.

Lets identify the missing interfaces and handle those correctly instead.

jsoverson:
In phantom's case the issue is alleviated by checking in HTMLMediaElement.js and HTMLAudioElement.js. To fully account for elements that don't exist, it would need to be managed everywhere.

If you're ok with the media subset I will change the PR to reflect managing those two cases.

arv:
Yes. Lets add the following:

if (OriginalHTMLMediaElement) return;

and same for HTMLAudioElement

jsoverson:
Updated the branch.

arv:
Perfect.

Did you sign the CLA?

@ebidel FYI

jsoverson:
Yes, I submitted the CLA this morning.

arv:
Great. I see it now.

jsoverson:
thanks @arv

PR #491, opened 2014-08-13
Make sure we set the constructor for wrapper classes

### LINKED ISSUE #139: Incorrect constructor for HTMLTemplateElement?
On chrome stable, `document.createElement('template').constructor == HTMLTemplateElement` is true while it is not on latest firefox (23.0.1), both using polymer polyfill.
Is this something I can rely on?

Note that on both browsers `document.createElement('template') instanceof HTMLTemplateElement` and `document.createElement('div') == HTMLDivElement` are true.

arv:
This adds tests for all the classes we use.

Fixes https://github.com/Polymer/TemplateBinding/issues/139

### DIFF
--- a/src/MutationObserver.js
+++ b/src/MutationObserver.js
@@ -255,6 +255,8 @@
   }
 
   MutationObserver.prototype = {
+    constructor: MutationObserver,
+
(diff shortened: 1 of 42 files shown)


jmesserly:
lgtm!

PR #492, opened 2014-08-14
Add support for /deep/ to querySelector and querySelectorAll

arv:
Fixes #436


jmesserly:
lgtm! out of curiosity, what wa the story behind the skipped tests?

arv:
The skipped tests are due to the way we filter query results instead of walking the logical dom. With the filtering approach we fail when the element is not part of the composed tree.

I realized that we did not need to skip the entire tests and some of the skips were fixed in a follow up pr.

PR #496, opened 2014-08-27
allow createShadowRoot on body

### LINKED ISSUE #421: Content inside `<template>` breaks extending `<body>` element.
I wrote a custom element that extends `<body>` (http://jsbin.com/vohoc/1/edit). It worked perfectly fine in both Canary and regular Chrome (http://jsbin.com/cepew/2/quiet).

But when I put anything into element's template:

``` html
<polymer-element name="x-foo" extends="body">
  <template>Something here breaks it.<content></content></template>
```

http://jsbin.com/cepew/3/quiet

It blows with a ton of `Uncaught Error: Assertion failed` errors. Two at the beginning at PointerGestureEvent.js:18, and throws one for every mouse move. 
(_Chrome 32.0.1700.107 m_)

jmesserly:
Fixes https://github.com/Polymer/polymer/issues/421

### DIFF
--- a/src/ShadowRenderer.js
+++ b/src/ShadowRenderer.js
@@ -73,7 +73,7 @@
       refChildWrapper.previousSibling_ = refChildWrapper.previousSibling;
     }
 
-    parentNode.insertBefore(newChild, refChild);
+    scope.originalInsertBefore.call(parentNode, newChild, refChild);
   }
 
   function remove(nodeWrapper) {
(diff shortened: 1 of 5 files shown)


src/ShadowRenderer.js
```diff
@@ -73,7 +73,7 @@
       refChildWrapper.previousSibling_ = refChildWrapper.previousSibling;
     }
 
-    parentNode.insertBefore(newChild, refChild);
+    scope.originalInsertBefore.call(parentNode, newChild, refChild);
```
jmesserly, line 76: note: the call through scope here is to support the "minimal DOM changes" test.

arv:
LGTM

PR #497, opened 2014-08-27
restore isLoadLikeEvent

jmesserly:
fixes https://github.com/Polymer/platform/issues/83

isLoadLikeEvent is from: https://github.com/Polymer/ShadowDOM/commit/b04881f3b00925daa96b6c7256fb046d9706d499#diff-960ceb36e2dc9470f018834bd9019745L190

I wrote the test first to verify it reproduces the issue (it passes if shadowdom.js+wrap call is not there, before the fix and with shadowdom.js, it fails at the 2000ms timeout)

### DIFF
--- a/src/wrappers/events.js
+++ b/src/wrappers/events.js
@@ -237,6 +237,19 @@
     }
   }
 
+
+  function isLoadLikeEvent(event) {
+    switch (event.type) {
+      // http://www.whatwg.org/specs/web-apps/current-work/multipage/webappapis.html#events-and-the-window-object
+      case 'load':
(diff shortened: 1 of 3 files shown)


jmesserly:
@arv mind taking a look?

kegluneq:
LGTM -- thanks!

arv:
LGTM

jmesserly:
cool, going to reland this in https://github.com/Polymer/ShadowDOM/pull/499, should be passing on IE/Safari now (_fingers crossed_)

PR #500, opened 2014-08-28
delete getDefaultComputedStyle from window

jakemac53:
Looks like this was overlooked in https://github.com/Polymer/ShadowDOM/commit/864ca98, polyfill does not take affect in firefox without it.


arv:
LGTM

PR #502, opened 2014-08-29
don't discard destination insertion points of the shadow host

### LINKED ISSUE #512: ShadowDOM renderer invalidated after insertion-point distribution
EDIT: fixed formatting/markdown

Thanks for your work on Polymer!

This problem exists in v0.2.4 but not v0.2.3 -- possible solution described below.

Given 3 web components A, B, and C: 

( see example https://gist.github.com/jcmoore/3432b06621e3c3ccd39c#file-shadow_events-html )

if A has changed-listerner and a corresponding getter on its prototype that will trigger a call to ShadowDOMPolyfill.renderAllPending(),

when C has a B in its shadow DOM template and C’s content-tag is inside of that B, and B has its own content-tag in its shadow DOM template,

when an A is placed in the light DOM of a C, B will not get events that originate on A.

This seems to be because during C's upgrade, its ShadowRenderer is invalidated and made pending before A's (and B's).  C will then (synchronously) begin upgrading its subtree (which includes A and B).  During A's upgrade, its createPropertyObserver() will trigger ShadowDOMPolyfill.renderAllPending() which will cause C's already queued ShadowRenderer to distribute insertion points (including B's content-tag) to A before A's own createShadowRoot() is called -- which will in turn cause its own ShadowRenderer to be invalidated and queued, ultimately abandoning the previously stored insertion points (especially those from B) at the next execution of ShadowDOMPolyfill.renderAllPending().

The result can be seen in ShadowDOM/src/wrappers/events.js' getEventPath() method, which will return a path missing B's insertion point(s) -- so event listeners on B will not fire for events on A.

I'm not sure what the side-effects are, but the undesirable invalidation/distribution order might be prevented by delaying any execution of ShadowDOMPolyfill.renderAllPending() until after any "currently upgrading" nodes have finished upgrading -- perhaps by making the following changes to ShadowDOM/src/ShadowRenderer.js:

https://gist.github.com/jcmoore/3432b06621e3c3ccd39c#file-shadowdom-src-shadowrenderer-js-L132-L156

and the following changes to CustomElements/src/CustomElements.js:

https://gist.github.com/jcmoore/3432b06621e3c3ccd39c#file-customelements-src-customelements-js-L235-L254

jmesserly:
when we rerender a shadow host, make sure not to discard information about where it was distributed itself (since its own renderer won't take care of that). fixes https://github.com/Polymer/polymer/issues/512

### DIFF
--- a/src/ShadowRenderer.js
+++ b/src/ShadowRenderer.js
@@ -298,7 +298,7 @@
 
     // http://w3c.github.io/webcomponents/spec/shadow/#distribution-algorithms
     distribution: function(root) {
-      this.resetAll(root);
+      this.resetAllSubtrees(root);
       this.distributionResolution(root);
     },
 
@@ -308,6 +308,10 @@
       else
         resetDestinationInsertionPoints(node);
 
+      this.resetAllSubtrees(node);
+    },
+
+    resetAllSubtrees: function(node) {
       for (var child = node.firstChild; child; child = child.nextSibling) {
         this.resetAll(child);
       }
--- a/test/js/reprojection.js
+++ b/test/js/reprojection.js
@@ -262,4 +262,34 @@ suite('Shadow DOM reprojection', function() {
     var content3 = sr3.appendChild(document.createElement('content'));
     assert.equal(getVisualInnerHtml(div), '<a><b><d></d></b></a>');
   });
+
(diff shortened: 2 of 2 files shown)


arv:
LGTM

jmesserly:
thank you!

PR #503, opened 2014-09-03
Fix for "TypeError: Argument 1 of Window.getComputedStyle does not implement interface Element."

nazar-pc:
I got that error on canvas element, and the reason was that because of event bubbling event comes to `document` where this method fails.
Did not tested for error `TypeError: Argument 1 of Window.getDefaultComputedStyle does not implement interface Element.` but it may fix it in some cases as well.

### DIFF
--- a/src/wrappers/Window.js
+++ b/src/wrappers/Window.js
@@ -59,7 +59,7 @@
   mixin(Window.prototype, {
     getComputedStyle: function(el, pseudo) {
       renderAllPending();
-      if (el instanceof HTMLDocument) {
+      if (!(el instanceof HTMLElement)) {
         return true;
       }
       return originalGetComputedStyle.call(unwrap(this), unwrapIfNeeded(el),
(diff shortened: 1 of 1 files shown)


arv:
This seems wrong. If `el` is not an Element we should throw.

Can you provide a test that shows what your desired behavior is.

src/wrappers/Window.js
```diff
@@ -59,6 +59,9 @@
   mixin(Window.prototype, {
     getComputedStyle: function(el, pseudo) {
       renderAllPending();
+      if (!(el instanceof HTMLElement)) {
+        return true;
```
arv, line 63: Wrong return type

arv:
Can you file a bug instead since it isn't clear what kind of issues you are having?

nazar-pc:
Sure, I'll try to make demo reproducing this bug as soon as I have enough time.
I'm using fabric.js library, error happens somewhere inside it, you can try it by yourself.

PR #505, opened 2014-09-09
fixes #504, avoids global list of all MutationObservers

jmesserly:
They're now added to the global list on-demand, and it is sorted by ID before delivering. Sounds like this is similar to the Blink implementation (https://github.com/Polymer/ShadowDOM/issues/504#issuecomment-55013158)

### DIFF
--- a/src/MutationObserver.js
+++ b/src/MutationObserver.js
@@ -15,7 +15,7 @@
   var globalMutationObservers = [];
   var isScheduled = false;
 
-  function scheduleCallback(observer) {
+  function scheduleCallback() {
     if (isScheduled)
       return;
     setEndOfMicrotask(notifyObservers);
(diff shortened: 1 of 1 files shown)


jmesserly:
@arv mind taking a look?

src/MutationObserver.js
```diff
@@ -180,13 +183,11 @@
         record.oldValue = associatedStrings[uid];
 
       // 8.
+      if (!observer.records_.length) {
+        scheduleCallback(observer);
+      }
       observer.records_.push(record);
```
arv, line 191: Is this reordering intentional?

arv, line 191: Never mind.

arv:
LGTM

Nice work around

arv:
Actually. Could we move the call to scheduleCallback out of the loop. No need to do that work more than once.

jmesserly:
> Actually. Could we move the call to scheduleCallback out of the loop. No need to do that work more than once.

Done.

Yeah was thinking it probably didn't matter much perf wise (because of the length check, and the isScheduled check inside scheduleCallback), and I liked having one less state variable and slightly better "encapsulation" of the global list (only schedule/notify touch it). But those are very minor things in the grand scheme of things. :)

jmesserly:
by the way, @arv do you usually prefer to squash into a single commit before submitting a pull req?

arv:
LGTM

Squashing is preferred. Less churn.

jmesserly:
thanks! will squash the follow up & land :)

PR #506, opened 2014-09-09
Add workaround for broken Safari 8 DOM

arv:
Safari's DOM is broken. It is returns invalid descriptors. We therefore
treat all properties as read-write even if they are really read only.

Fixes https://github.com/Polymer/platform/issues/66

### DIFF
--- a/src/wrappers.js
+++ b/src/wrappers.js
@@ -174,6 +174,15 @@ window.ShadowDOMPolyfill = {};
     }
   }
 
+  // Safari 8 exposes WebIDL attributes as an invalid accessor property. Its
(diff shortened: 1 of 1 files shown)


arv:
PTAL

jmesserly:
LGTM, maybe add a link to the bug? https://bugs.webkit.org/show_bug.cgi?id=49739

arv:
Added a comment explaining what the hell is going on and added a link to the webkit bug.

jmesserly:
nice comment :+1:

benjamin-chang:
:+1:

PR #509, opened 2014-09-17
Throw an Error if window.FormData is undefined.

geppy:
This would address #508 by throwing an error if FormData is undefined.

I'd be happy to write find/write a polyfill if it's preferred to address the issue that way.


arv:
Lets not throw an error. I created #510 which just continues on like we do for everything else that the browser does not support.

geppy:
Great, thanks.

PR #510, opened 2014-09-17
Guard FormData for IE9

arv:
Since old IE does not have FormData we need to not even try to wrap it.

Fixes #508


arv:
PTAL

addyosmani:
:+1:

jmesserly:
lgtm too!

PR #514, opened 2014-10-06
Fix for error in Firefox because of workaround for Safari

nazar-pc:
Fix for https://github.com/Polymer/platform/issues/94
RegExp based on user agent examples from http://myip.ms/view/comp_browsers/3458/Safari_8.html beause I do not have any available machine with Safari 8

### DIFF
--- a/src/wrappers.js
+++ b/src/wrappers.js
@@ -179,12 +179,8 @@ window.ShadowDOMPolyfill = {};
   // shape of the descriptor and make all properties read-write.
   // https://bugs.webkit.org/show_bug.cgi?id=49739
   var isBrokenSafari = function() {
-    // Whether this is Safari 8 or not
-    if (!/AppleWebKit.*Version\/8\.0.*Safari/i.test(navigator.userAgent)) {
-      return false;
-    }
     var descr = Object.getOwnPropertyDescriptor(Node.prototype, 'nodeType');
-    return !!descr && 'set' in descr;
+    return descr && !descr.get && !descr.set;
   }();
 
   function installProperty(source, target, allowMethod, opt_blacklist) {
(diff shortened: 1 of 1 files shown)


src/wrappers.js
```diff
@@ -179,6 +179,10 @@ window.ShadowDOMPolyfill = {};
   // shape of the descriptor and make all properties read-write.
   // https://bugs.webkit.org/show_bug.cgi?id=49739
   var isBrokenSafari = function() {
+    // Whether this is Safari 8 or not
+    if (!/AppleWebKit.*Version\/8\.0.*Safari/i.test(navigator.userAgent)) {
+      return false;
+    }
     var descr = Object.getOwnPropertyDescriptor(Node.prototype, 'nodeType');
     return !!descr && 'set' in descr;
```
arv: How about checking that both `descr.set === undefined` and `descr.get === undefined` instead?

arv: Or simply

``` js
return descr && !descr.get && !descr.set;
```

nazar-pc: Yes, it may work. And will be an elegant fix.
But I'm unable to check it in real Safari 8.
Just checked in Firefox Nightly - your variant works fine.

arv: The problem with Safari (and WebKit nightlies) is that it returns `{get: undefined, set: undefined, ...}`. That is an invalid descriptor if there is no `value` property so maybe we need to ensure that there is no value as well?

nazar-pc: Not sure, I think it is better to test.
Here is updated example with `return descr && !descr.get && !descr.set;` fix instead of previous: http://jsfiddle.net/znq76pg9/1/

arv: My concern was that a browser might do some other crazy thing where it implements things as magic data properties.

Still. LGTM with `return descr && !descr.get && !descr.set;`. Care to update the PR?

nazar-pc: Updated

arv:
LGTM

Thanks.

PR #515, opened 2014-10-07
Fix textContent setter

jakemac53:
Previously doing myElement.textContent = null would result in a textContent == 'null'. It looks like the standard behavior for this should be to treat null, undefined, and empty string all as empty string.

### DIFF
--- a/src/wrappers/Node.js
+++ b/src/wrappers/Node.js
@@ -628,6 +628,7 @@
       return s;
     },
     set textContent(textContent) {
+      if (textContent == null) textContent = '';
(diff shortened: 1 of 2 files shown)


jmesserly:
great fix! fyi, this is spec'd at https://dom.spec.whatwg.org/#dom-node-textcontent

would be good to have a small unit test, perhaps add to https://github.com/Polymer/ShadowDOM/blob/master/test/js/Node.js?

jakemac53:
test added, ptal

jmesserly:
lgtm, thanks!

